PR #179, opened 2018-05-18
WP-387

paveldhq:


inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -147,10 +147,65 @@ public function getXMLFiltered(SubmissionEntity $submission)
...
+                if (array_key_exists($_fieldName, $targetContent)) {
+                    $lockedData['entity'][$_fieldName] = $targetContent[$_fieldName];
+                }
+            }
+        }
+
+        return $lockedData;
+    }
+
+    private function arrayMergeIfKeyNotExists($lockedData, $translation)
```
PavelLoparev: Seems like this method should be in `ArrayHelper` class. Can be static.

inc/Smartling/Extensions/TranslateLock.php
```diff
@@ -17,6 +21,79 @@
...
+    {
+        $this->contentHelper = $contentHelper;
+    }
+
+    /**
+     * @return LoggerInterface
+     */
+    public function getLogger()
+    {
+        if (!($this->logger instanceof LoggerInterface)) {
```
PavelLoparev: How is it possible to have logger with a type != `LoggerInterface`?

inc/Smartling/Extensions/TranslateLock.php
```diff
@@ -26,22 +103,59 @@ public function getName()
     }
 
     /**
-     * @inheritdoc
+     * @param $id
+     *
+     * @return SubmissionEntity|null
      */
-    public function register()
+    private function getSubmissionById($id)
```
PavelLoparev: Seems like this is some general method and shouldn't be in this class. Maybe we need some `SubmissionHelper`?

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -147,10 +147,65 @@ public function getXMLFiltered(SubmissionEntity $submission)
         }
     }
 
+    /**
+     * @param SubmissionEntity $submission
+     *
+     * @return array
+     */
+    private function readLockedTranslationFieldsBySubmission(SubmissionEntity $submission)
```
PavelLoparev: Trait `SmartlingCoreUploadTrait` is getting overflown by such unrelated methods. I think it should be in some helper class of directly in your new `TranslateLock` class.

inc/Smartling/Base/SmartlingCoreUploadTrait.php
paveldhq: agree for static, but it should not be inside ArrayHelper

inc/Smartling/Extensions/TranslateLock.php
paveldhq: don't think so

PR #181, opened 2018-12-18
Wp 434: Reimplement how we read CDATA section

paveldhq:
Careful read of CDATA section of XML

### DIFF
--- a/tests/Smartling/Helpers/ShortcodeHelperTest.php
+++ b/tests/Smartling/Helpers/ShortcodeHelperTest.php
@@ -1,0 +1,68 @@
+<?php
+
+namespace Smartling\Tests\Smartling\Helpers;
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Class XmlEncoder
+ * @package Smartling\Tests\Smartling\Helpers
+ * @covers  \Smartling\Helpers\ShortcodeHelper
+ */
+class ShortcodeHelperTest extends TestCase
+{
+    /**
+     * @covers       \Smartling\Helpers\ShortcodeHelper::getCdata
+     * @dataProvider getCDATADataProvider
+     *
+     * @param string   $expectedString
+     * @param \DOMNode $node
+     */
+    public function testGetCdata($expectedString, \DOMNode $node)
+    {
+        self::assertEquals($expectedString, \Smartling\Helpers\ShortcodeHelper::getCdata($node));
+    }
+
+    /**
+     * Creates \DOMNode with children CDATA with content from $parts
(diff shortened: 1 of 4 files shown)


tests/Smartling/Helpers/ShortcodeHelperTest.php
```diff
@@ -0,0 +1,68 @@
...
+        return [
+            implode('', $parts),
+            $this->makeDomNode($parts),
+        ];
+    }
+
+    public function getCDATADataProvider()
+    {
+        $sets = [
+            ['a', 'b', 'c'],
```
PavelLoparev: In case of `['strangeb', 'roken string']` I agree, resulting string must be `strange broken string` (direct concatenation). But in case of `['a', 'b', 'c']` I'm not sure about `abc`, I mean `['Hello', 'World']` will be `HelloWorld` in your case.

@dimitrystd do you know what rules are used by parsers to break strings? I'm 99% sure they don't break words.

tests/Smartling/Helpers/ShortcodeHelperTest.php
```diff
@@ -0,0 +1,68 @@
...
+            ['strangeb', 'roken string'],
+            ['one element'],
```
PavelLoparev: In case of `['strangeb', 'roken string']` I agree, resulting string must be `strange broken string` (direct concatenation). But in case of `['a', 'b', 'c']` I'm not sure about `abc`, I mean `['Hello', 'World']` will be `HelloWorld` in your case.

@dimitrystd do you know what rules are used by parsers to break strings? I'm 99% sure they don't break words.

paveldhq: that is true, result a concatination of CDATA children.
The most common case of multiple CDATA children - content that may be processed as a closing tag os CDATA block. For our purpose we need just the whole content from all CDATA sections.

PavelLoparev: You mean this case `['<tag>', 'Some strings', '</tag>'] -> '<tag>Some string</tag>'`?

paveldhq: yes, but in our case it can be `['<','tag>Some string<','/tag>']`

dimitrystd: I found the ticket about cdata https://bt.smartling.net/browse/SLDSS-18056. I was wrong about splitting single block. It seems it's another use case.
I'm ok to concat without space.

PR #183, opened 2019-01-14
Sovos file issue coverage for

paveldhq:
covers \Smartling\Helpers\FieldsFilterHelper::structurizeArray

### DIFF
--- a/tests/Smartling/Helpers/FieldsFilterHelperTest.php
+++ b/tests/Smartling/Helpers/FieldsFilterHelperTest.php
@@ -1,0 +1,111 @@
+<?php
+
+namespace Smartling\Tests\Smartling\Helpers;
+
+use PHPUnit\Framework\TestCase;
+use Smartling\Helpers\FieldsFilterHelper;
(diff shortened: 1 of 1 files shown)


tests/Smartling/Helpers/FieldsFilterHelperTest.php
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * @return array
+     */
+    public function structurizeArrayDataProvider()
+    {
+        $fields = [
+            'a'                                                                   => ['a' => 'test'],
+            ''                                                                    => ['' => 'test'],
```
dimitrystd: Is it possible to pass `null` here?

paveldhq: no, it is not possible due to type hinting in original method that is tested.
first argument should be `array`, otherwise it will fail with something like:
`TypeError : Argument 1 passed to Smartling\Helpers\FieldsFilterHelper::structurizeArray() must be of the type array, null given`

PR #185, opened 2019-01-31
WP-443. PHP 7.3 deprecation warnings

paveldhq:


inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
```diff
@@ -372,7 +372,7 @@ private function verifyDefinitions(array $localDefinitions, array $dbDefinitions
                         ) {
                             // ACF Option Pages has internal issue in definition, so skip it:
                             if ('group_572b269b668a4' === $local['parent']) {
-                                continue;
+                                break;
```
PavelLoparev: Seems like you don't even need `switch` here as you check only one case. I think `if/else` would be better fit here.

PavelLoparev:
Please, write unit tests before making changes. Then make these changes and confirm that everything work as expected.

inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
paveldhq: created follow-up task

PR #186, opened 2019-02-12
Gutenberg blocks support

paveldhq:


inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,352 @@
...
+    public function renderTranslatedBlockNode(\DOMNode $node)
+    {
+        $blockName = $node->getAttribute('blockName');
+        $blockName = '' === $blockName ? null : $blockName;
+        $originalAttributesEncoded = $node->getAttribute('originalAttributes');
+        $chunks = [];
+        $translatedAttributes = [];
+        $this->sortChildNodesContent($node, $chunks, $translatedAttributes);
+        // simple plain blocks
+        if (null === $blockName) {
```
PavelLoparev: Why do we need the `$blockName = '' === $blockName ? null : $blockName;` line? Seems like `if` statement can be just `if (empty($blockName)) {`

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,352 @@
...
+            return implode('\n', $chunks);
```
PavelLoparev: Again, it's not so easy to understand why you implode an empty array here (hint, because `sortChildNodesContent` modifies passed `$chunks` and `$translatedAttributes` variables).

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,352 @@
...
+     * @param string $name
+     * @param array  $attrs
+     * @param array  $chunks
+     * @return string
+     */
+    private function renderGutenbergBlock($name, array $attrs = [], array $chunks = [])
+    {
+        $attributes = 0 < count($attrs) ? ' ' . json_encode($attrs) : '';
+        $content = implode('', $chunks);
+        return ('' !== $content)
```
PavelLoparev: `return (!empty($content))`?

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,352 @@
...
+     *
+     * @return TranslationStringFilterParameters
+     */
+    public function processTranslation(TranslationStringFilterParameters $params)
+    {
+        $this->setParams($params);
+        $node = $this->getNode();
+        $string = static::getCdata($node);
+
+        if ('' === $string) {
```
PavelLoparev: `if (empty($string)) {`?

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,352 @@
...
+            foreach ($children as $child) {
+                /**
+                 * @var \DOMNode $child
+                 */
+                if (static::BLOCK_NODE_NAME === $child->nodeName) {
+                    $string .= $this->renderTranslatedBlockNode($child);
+                }
+            }
+
+            foreach ($children as $child) {
```
PavelLoparev: Can we collapse these two loops into one?

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    /**
+     * @throws SmartlingGutenbergNotFoundException
+     */
+    private function loadBlockClass()
+    {
+        $paths = [
+            vsprintf('%swp-includes/blocks.php', [ABSPATH]),
+            vsprintf('%swp-content/plugins/gutenberg/lib/blocks.php', [ABSPATH]),
```
PavelLoparev: I would like to make it dynamic. I mean seems like there should be a hook where you can tell the plugin where to look for guttenberg file and then merge default paths with retrieved from user defined hooks. I can imagine situation when in WP5.x.x these locations will be changed.

inc/Smartling/Helpers/SubstringProcessorHelperAbstract.php
```diff
@@ -0,0 +1,329 @@
...
+    /**
+     * @param string $blockName
+     *
+     * @return array|false
+     */
+    public function getBlockAttributes($blockName)
+    {
+        return array_key_exists($blockName, $this->blockAttributes)
+            ? $this->blockAttributes[$blockName]
+            : false;
```
PavelLoparev: I think this should return an empty array instead of false.

composer.json
```diff
@@ -8,7 +8,8 @@
     "symfony/dependency-injection" : "~2.8",
     "symfony/config" :               "~2.8",
     "symfony/yaml" :                 "~2.8",
-    "smartling/api-sdk-php" :        "~3.2"
+    "smartling/api-sdk-php" :        "~3.2",
+    "ext-dom": "*"
```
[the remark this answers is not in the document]
paveldhq: adding `"ext-dom": "*"` removes PHPStorm `composer.json` warning

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,352 @@
...
+        return $attributes;
+    }
+
+    /**
+     * @param $string
+     * @return bool
+     */
+    private function hasBlocks($string)
+    {
+        return (false !== strpos($string, '<!-- wp:'));
```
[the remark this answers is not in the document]
paveldhq: yes, see original spec:
https://wordpress.org/gutenberg/handbook/designers-developers/key-concepts/#the-anatomy-of-a-serialized-block

and own gutenberg's code:

https://github.com/WordPress/gutenberg/blob/40872d38af36b09f64245d986e039ec7a692eee1/lib/blocks.php#L42-L45

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,352 @@
...
+
+        return $params;
+    }
+
+    /**
+     * @param \DOMNode $node
+     * @param array    $chunks
+     * @param array    $attrs
+     */
+    private function sortChildNodesContent(\DOMNode $node, array & $chunks, array & $attrs)
```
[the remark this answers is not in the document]
paveldhq: 1. it sorts child nodes between 2 types:
 - chunk
 - attribute
2. removes extracted sorted data from original scope.

Sort != order

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,352 @@
...
+        if (0 < count($originalAttributes)) {
+            $flatAttributes = $this->getFieldsFilter()->flatternArray($originalAttributes);
+
+            if (0 < count($flatAttributes)) {
+                $attr = self::maskAttributes($blockName, $flatAttributes);
+                $attr = $this->postReceiveFiltering($attr);
+                $attr = self::unmaskAttributes($blockName, $attr);
+            }
+
+            $filteredAttributes = array_merge($flatAttributes, $attr, $translatedAttributes);
```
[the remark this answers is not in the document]
paveldhq: refactored

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,352 @@
...
+
+            $processedAttributes = $this->getFieldsFilter()->structurizeArray($filteredAttributes);
+        } else {
```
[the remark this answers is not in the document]
paveldhq: refactored

composer.json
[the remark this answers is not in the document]
PavelLoparev: Could you please post here `composer.json` warning?

inc/Smartling/Helpers/GutenbergBlockHelper.php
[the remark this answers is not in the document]
PavelLoparev: Should we use regexp `<!--\s+wp` like in guttenber parser code?

inc/Smartling/Helpers/GutenbergBlockHelper.php
paveldhq: `null` is used by original WP, so just following original WP namings.
`''` is produced when saving `null` to XML Node Attribute value, here also just following original  way created by WP

inc/Smartling/Helpers/GutenbergBlockHelper.php
paveldhq: Array is not empty here.
It is filled with data by `sortChildNodesContent()`

inc/Smartling/Helpers/GutenbergBlockHelper.php
paveldhq: what advantages does `empty` have in comparison to strict typed comparison?

inc/Smartling/Helpers/GutenbergBlockHelper.php
paveldhq: what advantages does `empty` have in comparison to strict typed comparison?

inc/Smartling/Helpers/GutenbergBlockHelper.php
paveldhq: Due to implementation of `DOM` php extension we can't.
Internal references to `current()` work not as expected with `foreach()` when removing elements.

inc/Smartling/Helpers/GutenbergBlockHelper.php
paveldhq: First variant is for native WP5, second is for WP < 5 + gutenberg plugin.
There is no API to get information you mean, also note that Gutenberg works on Frontend (PHP renderer) and Admin panel (JS part). Our plugin works ONLY in Admin panel (Backend) and Cron, so normally we don't intersect with gutenberg at all.

inc/Smartling/Helpers/SubstringProcessorHelperAbstract.php
paveldhq: Why? Logical function return not boolean value?

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -405,7 +406,7 @@ public function bulkSubmit(SubmissionEntity $submission)
             $locales = [];
 
             $xml = $this->getXMLFiltered($submission);
-
+;
```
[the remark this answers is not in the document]
paveldhq: removed

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -428,6 +429,7 @@ public function bulkSubmit(SubmissionEntity $submission)
             }
 
             if (!StringHelper::isNullOrEmpty($xml)) {
+                //Bootstrap::DebugPrint($xml, true);
```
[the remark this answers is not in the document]
paveldhq: removed

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,352 @@
...
+     */
+    private function loadBlockClass()
+    {
+        $paths = [
+            vsprintf('%swp-includes/blocks.php', [ABSPATH]),
+            vsprintf('%swp-content/plugins/gutenberg/lib/blocks.php', [ABSPATH]),
+        ];
+
+        foreach ($paths as $path) {
+            //$this->getLogger()->debug(vsprintf('Trying to get block class from file: %s', [$path]));
```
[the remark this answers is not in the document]
paveldhq: Don't think so. Removed.
have special `SmartlingGutenbergNotFoundException`

composer.json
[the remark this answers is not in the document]
PavelLoparev: I think it might be related to `php7-xml` package

inc/Smartling/Helpers/GutenbergBlockHelper.php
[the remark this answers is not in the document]
paveldhq: `strpos` is faster than `preg_match`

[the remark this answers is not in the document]
paveldhq: Gutenberg's function:
```php
if ( ! function_exists( 'has_blocks' ) ) {
	/**
	 * Determine whether a post or content string has blocks.
	 *
	 * This test optimizes for performance rather than strict accuracy, detecting
	 * the pattern of a block but not validating its structure. For strict accuracy
	 * you should use the block parser on post content.
	 *
	 * @since 3.6.0
	 * @see gutenberg_parse_blocks()
	 *
	 * @param int|string|WP_Post|null $post Optional. Post content, post ID, or post object. Defaults to global $post.
	 * @return bool Whether the post has blocks.
	 */
	function has_blocks( $post = null ) {
		if ( ! is_string( $post ) ) {
			$wp_post = get_post( $post );
			if ( $wp_post instanceof WP_Post ) {
				$post = $wp_post->post_content;
			}
		}

		return false !== strpos( (string) $post, '<!-- wp:' );
	}
}
```
So since i have not anything but a string i can just use `strpos`

composer.json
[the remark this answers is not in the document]
PavelLoparev: https://stackoverflow.com/questions/43408604/php7-install-ext-dom-issue

inc/Smartling/Helpers/GutenbergBlockHelper.php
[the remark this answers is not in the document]
PavelLoparev: Then just remove it

inc/Smartling/Helpers/GutenbergBlockHelper.php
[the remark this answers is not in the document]
PavelLoparev: got it.

inc/Smartling/Helpers/GutenbergBlockHelper.php
PavelLoparev: got it

inc/Smartling/Helpers/SubstringProcessorHelperAbstract.php
PavelLoparev: > Logical function return not boolean value?

`@return array|false` so the same question for you) moreover, `getBlockAttributes` doesn't seem to be logical function (which usually start from `is...`).

But anyway, it's not blocker/critical issue.

paveldhq: done

inc/Smartling/Helpers/GutenbergBlockHelper.php
PavelLoparev: Readability. But I think you can leave as is as it's not critical.

inc/Smartling/Helpers/GutenbergBlockHelper.php
PavelLoparev: Readability. But I think you can leave as is as it's not critical.

inc/Smartling/Helpers/GutenbergBlockHelper.php
PavelLoparev: got it

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,359 @@
...
+            ],
+            ''
+        );
+
+        foreach ($block['innerContent'] as $chunk) {
+            $part = null;
+            if (is_string($chunk)) {
+                $part = $this->createDomNode(static::CHUNK_NODE_NAME, [], $chunk);
+            } else {
+                $part = $this->placeBlock($block['innerBlocks'][$indexPointer++]);
```
dimitrystd: It's hard to read\understand why do we use `$indexPointer` here.
We enumerate `$block['innerContent']` but in some cases we take data from completely different place `$block['innerBlocks']`. It would be nice to have wiki page with gutenberg html vs smartling xml. Then it will be more obvious how we get `gutenbergBlock`; `contentChunk` and `blockAttribute` in xml.

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,359 @@
...
+
+        return $params;
+    }
+
+    /**
+     * @param \DOMNode $node
+     * @param array    $chunks
+     * @param array    $attrs
+     */
+    public function sortChildNodesContent(\DOMNode $node)
```
dimitrystd: I didn't get what we do in this method. I see we populate 2 arrays `$chunks` and `$attrs` but never use them inside method and don't return outside.

dimitrystd: I found the comments below. Let's get rid this magic.

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -0,0 +1,359 @@
...
+        if (!$this->hasBlocks($string)) {
+            return $params;
+        }
+        $originalBlocks = gutenberg_parse_blocks($string);
+
+        foreach ($originalBlocks as $block) {
+            $node = $this->placeBlock($block);
+            $params->getNode()->appendChild($node);
+        }
+        self::replaceCData($params->getNode(), '');
```
dimitrystd: What is the reason to remove content in CDATA section?

inc/Smartling/Helpers/GutenbergBlockHelper.php
paveldhq: Agree, but that is not our idea. That ugly logic is a part of original Wordpress `render_block()`  function: https://github.com/WordPress/WordPress/blob/master/wp-includes/blocks.php#L203-L211 .
We cannot use it as it is because of internal `apply_filter()` calls.
As argument it receives a result of `parse_blocks()` https://github.com/WordPress/WordPress/blob/master/wp-includes/blocks.php#L241-L253 .
and the result is an `array` with quite strange structure.


That is a part of standard Wordpress logic:
https://github.com/WordPress/WordPress/blob/master/wp-includes/blocks.php#L203-L211

inc/Smartling/Helpers/GutenbergBlockHelper.php
paveldhq: we remove content because it is divided into substrings and stored as child nodes.

inc/Smartling/Helpers/GutenbergBlockHelper.php
paveldhq: done

PR #190, opened 2019-03-13
Handle unexpected 0 response from wp_insert_post

paveldhq:


dimitrystd:
New code looks even better then previous

PR #199, opened 2019-05-17
Wp 466

paveldhq:


inc/Smartling/ContentTypes/AutoDiscover/Taxonomies.php
```diff
@@ -17,6 +17,11 @@ class Taxonomies
      */
     private $logger;
 
+    /**
+     * @var array
+     */
+    private $ignoredTypes;
```
dimitrystd: Probably it's time to extract common properties in base class. I'm talking about `di`, `logger`, `ignoredTypes `

PR #200, opened 2019-05-31
fix: remove SCRIPT_DEBUG

modelm:
To avoid throwing PHP notices when other plugins or config already define `SCRIPT_DEBUG`, its definition should be tested before calling `define()`.

I disagree with hardcoding this to true if it's not already explicitly defined on live sites as well - is there a good reason to force worse performance on users?

### DIFF
--- a/smartling-connector.php
+++ b/smartling-connector.php
@@ -21,7 +21,9 @@
 if (!defined('WPINC')) {
     die;
 }
-define('SCRIPT_DEBUG', true);
+if (!defined('SCRIPT_DEBUG')) {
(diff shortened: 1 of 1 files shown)


smartling-connector.php
```diff
@@ -21,7 +21,9 @@
 if (!defined('WPINC')) {
     die;
 }
-define('SCRIPT_DEBUG', true);
+if (!defined('SCRIPT_DEBUG')) {
```
paveldhq: I think we can remove it, since we're nor using it anymore.

modelm: Cool, thanks!

PR #203, opened 2019-06-11
WP-470

paveldhq:


PavelLoparev:
Forgot to mention: unit tests required for catching issues like this.

PR #207, opened 2019-06-13
Do not use hardcoded default content-type on BulkSubmit page

paveldhq:


inc/Smartling/WP/Table/BulkSubmitTableWidget.php
```diff
@@ -129,6 +130,12 @@ public function __construct(
 
         $this->setLogger($entityHelper->getLogger());
 
+        /*
+         * Set default content type to first known
+         */
+        $this->defaultValues[static::CONTENT_TYPE_SELECT_ELEMENT_NAME] =
```
dimitrystd: It seems we can clean up code in this class above
```
    private $defaultValues = [
        self::CONTENT_TYPE_SELECT_ELEMENT_NAME  => 'post',
...
```
Remove `post` that may confuse someone who is not familiar with this code.

PR #208, opened 2019-07-17
WP-478

paveldhq:
Splitting filtering into:
 - content check
 - filtering

### DIFF
--- a/inc/Smartling/Helpers/ContentHelper.php
+++ b/inc/Smartling/Helpers/ContentHelper.php
@@ -344,6 +344,7 @@ class ContentHelper
     public function checkEntityExists($blogId, $contentType, $contentId)
     {
         $needSiteSwitch = (int)$blogId !== $this->getSiteHelper()->getCurrentBlogId();
+        $result         = false;
 
         if ($needSiteSwitch) {
             $this->getSiteHelper()->switchBlogId((int)$blogId);
@@ -351,7 +352,7 @@ class ContentHelper
 
         try {
             if (($this->getIoFactory()->getMapper($contentType)->get($contentId)) instanceof EntityAbstract) {
-                return true;
+                $result = true;
             }
         } catch (Exception $e) {
         }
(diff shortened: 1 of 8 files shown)


inc/Smartling/Helpers/ContentHelper.php
```diff
@@ -339,4 +340,25 @@ public function removeTargetMetadata(SubmissionEntity $submission)
...
+        if ($needSiteSwitch) {
+            $this->getSiteHelper()->switchBlogId((int)$blogId);
+        }
+
+        try {
+            if (($this->getIoFactory()->getMapper($contentType)->get($contentId)) instanceof EntityAbstract) {
+                return true;
+            }
+        } catch (Exception $e) {
+        }
```
PavelLoparev: Maybe log `warn` or something here?

inc/Smartling/Helpers/MetaFieldProcessor/ReferencedStdBasedContentProcessorAbstract.php
```diff
@@ -92,10 +92,20 @@ public function processFieldPostTranslation(SubmissionEntity $submission, $field
                 )
             );
 
-            $contentType = null;
-            try {
-                $contentType = $this->detectRealContentType($submission->getSourceBlogId(), $value);
-            } catch (SmartlingWpDataIntegrityException $e) {
+            $contentType = $this->detectRealContentType($submission->getSourceBlogId(), $value);
```
PavelLoparev: According to previous code `detectRealContentType` might throw `SmartlingWpDataIntegrityException` exception but now you don't catch this error. Morever, now you log `Couldn\'t identify content type for id=\'%s\', blog=\'%s\'. Keeping existing value \'%s\'.` error message when `checkEntityExists` returns `false` which doesn't seem right. This error should be logged when `detectRealContentType` throws exception.

paveldhq: The code was simplified since we need to go through several steps to read content and we don't need to know which step stops reading we just need to get content or any `Exception`.

PR #209, opened 2019-07-26
WP-485 Virtual shortcode support broken.

paveldhq:
due to update in `wp-include/shortcodes.php:140` `function has_shortcode()` added additional check that now requires rewrite of original registered shortcodes to run.

### DIFF
--- a/inc/Smartling/Helpers/ShortcodeHelper.php
+++ b/inc/Smartling/Helpers/ShortcodeHelper.php
@@ -94,16 +94,23 @@ class ShortcodeHelper extends SubstringProcessorHelperAbstract
     private function hasShortcodes($string)
     {
         $possibleShortcodes = $this->getRegisteredShortcodes();
+
+        global $shortcode_tags;
+        $oldTags = $shortcode_tags;
+
+        $shortcode_tags = array_flip($possibleShortcodes);
+
         foreach ($possibleShortcodes as $possibleShortcode) {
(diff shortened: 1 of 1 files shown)


inc/Smartling/Helpers/ShortcodeHelper.php
```diff
@@ -94,16 +94,23 @@ public function register()
     private function hasShortcodes($string)
     {
         $possibleShortcodes = $this->getRegisteredShortcodes();
+
+        global $shortcode_tags;
+        $oldTags = $shortcode_tags;
+
+        $shortcode_tags = array_flip($possibleShortcodes);
```
PavelLoparev: I don't understand where "new" `$shortcode_tags` are being used.

inc/Smartling/Helpers/ShortcodeHelper.php
```diff
@@ -150,7 +155,6 @@ public function processTranslation(TranslationStringFilterParameters $params)
                 }
             }
 
-            $node->appendChild(new \DOMCdataSection($string));
```
PavelLoparev: Why this logic is changed? Why don't we need to add child node now?

paveldhq: it is useless since we replace it then with a processed string (line 167):
`static::replaceCData($node, $string_m);`

PR #211, opened 2019-08-06
Wp 485 replace shortcode parser

paveldhq:


inc/Smartling/Helpers/ShortcodeHelper.php
```diff
@@ -344,11 +342,16 @@ private static function buildShortcodeAttributes(array $attributes = [])
     {
         $attributesString = '';
 
+        $isIneger = function ($data) {
+            return is_int($data)
+                || (string)(int)$data === $data;
```
PavelLoparev: Why do you need extra check after PHP's core `is_int`? What does this extra check actually do?

inc/Smartling/Helpers/ShortcodeHelper.php
```diff
@@ -360,6 +363,15 @@ private static function buildShortcodeAttributes(array $attributes = [])
         return $attributesString;
     }
 
+    private static function escapeValue($data)
+    {
+        if (function_exists('esc_attr')) {
```
PavelLoparev: Does Wordpress core provide this function? In which cases it might be missing?

inc/Smartling/Helpers/SubstringProcessorHelperAbstract.php
```diff
@@ -134,7 +134,7 @@ protected function nodeToArray(\DOMNode $node)
      */
     public function addBlockAttribute($blockName, $attributeName, $translatedString, $originalHash)
     {
-        $this->blockAttributes[$blockName][$attributeName] = [$originalHash => $translatedString];
+        $this->blockAttributes[$blockName][$attributeName][$originalHash] = $translatedString;
```
PavelLoparev: isn't it the same as it was before?

inc/Smartling/Helpers/ShortcodeHelper.php
```diff
@@ -114,6 +112,52 @@ private function hasShortcodes($string)
...
+
+        /**
+         * Walking back to avoid internal pointer reset
+         */
+        $index = $node->childNodes->count();
+        if (0 === $index) {
+            return $translations;
+        }
+        while ($index) {
+            $cNode = $node->childNodes->item(--$index);
```
PavelLoparev: Just comment, no action required. This is not `index`, this is count. If it's `index`, then you should've done `$index--` instead of `--$index`.

inc/Smartling/Helpers/SubstringProcessorHelperAbstract.php
```diff
@@ -122,7 +122,7 @@ protected function nodeToArray(\DOMNode $node)
              */
             $struct[$attributeName] = $attributeValue->value;
         }
-        $struct['value'] = $node->nodeValue;
+        $struct['value'] = static::getCdata($node);
```
PavelLoparev: Does it mean that all previous connector versions worked with raw `nodeValue` instead of value inside `cdata`? What consequences does it produce?

inc/Smartling/Helpers/ShortcodeHelper.php
paveldhq: seems better name is `$internalPointer`, and i'm going from `count` to `0`, so it should be `--$internalPointer`

inc/Smartling/Helpers/ShortcodeHelper.php
paveldhq: `is_int("6")` returns `false`, so need to check if string contains an integer number

inc/Smartling/Helpers/ShortcodeHelper.php
paveldhq: it is missing while running unit tests, yes, that is a WP core function and it does several hooks under the hood, so there is no easy way to mock it.

inc/Smartling/Helpers/SubstringProcessorHelperAbstract.php
paveldhq: `static::getCdata($node);` is more correct in case that there might be another child nodes since our API is open

inc/Smartling/Helpers/SubstringProcessorHelperAbstract.php
paveldhq: that is a logic issue when we have several translated attributes with same name of shortcode and same name of attribute name but different content. Discovered by unit test.

inc/Smartling/Helpers/ShortcodeHelper.php
PavelLoparev: So, then tests are testing something else because your `"mock"` function implementation differs from original `esc_attr`

inc/Smartling/Helpers/SubstringProcessorHelperAbstract.php
PavelLoparev: Sounds like you meant `$this->blockAttributes[$blockName][$attributeName][] = [$originalHash => $translatedString];`

inc/Smartling/Helpers/ShortcodeHelper.php
PavelLoparev: mm...`is_numeric`? https://www.php.net/manual/en/function.is-numeric.php

paveldhq: is_numeric will return 'true' also to float type, but i need only check only for integer values

inc/Smartling/Helpers/ShortcodeHelper.php
paveldhq: i don't test esc_attr, i test own code.

PavelLoparev: omg. `is_numeric($var) && is_int($var)`?

inc/Smartling/Helpers/ShortcodeHelper.php
PavelLoparev: omg. `is_numeric($var) && is_int($var)`?)

inc/Smartling/Helpers/ShortcodeHelper.php
PavelLoparev: > i don't test esc_attr, i test own code.

If you mocked this function then it's used somewhere by your code. And in your test you assert result of `htmlentities()` instead of what Wordpress does.

PavelLoparev: Oh, tests just checks that there will be no quote and no difference who will change quote to something else. Ok

PavelLoparev:
Failing build

PR #212, opened 2019-08-28
Wp 480 detect relation images in html links

paveldhq:


inc/Smartling/Services/ContentRelationDiscoveryService.php
```diff
@@ -0,0 +1,281 @@
...
+
+    public function actionHandler()
+    {
+        $contentType = $this->getContentType();
+        $id          = $this->getId();
+        $curBlogId   = $this->getContentHelper()->getSiteHelper()->getCurrentBlogId();
+
+        if ($this->getContentHelper()->checkEntityExists($curBlogId, $contentType, $id)) {
+
+            $ioWrapper = $this->getContentHelper()->getIoFactory()->getMapper($contentType);
```
PavelLoparev: `ioWrapper` or `ioMapper`?

inc/Smartling/Services/ContentRelationDiscoveryService.php
```diff
@@ -0,0 +1,281 @@
...
+            $this->returnSuccess(
+                [
+                    'items' => [
+                        $detectedReferences,
+                    ],
+                ]
+            );
+
+        } else {
+            $this->returnError('content.not.exists', 'Requested content is not found', 404);
```
PavelLoparev: maybe `content.not.found`

inc/Smartling/Services/ContentRelationDiscoveryService.php
paveldhq: Method is implemented in the base class, but in that case I'm getting the wrapper to read content, so its ok.

PR #214, opened 2019-09-01
Wp 477 yext related items backend

paveldhq:


inc/Smartling/Services/ContentRelationDiscoveryService.php
```diff
@@ -0,0 +1,577 @@
...
+
+    /**
+     * @var SubmissionManager
+     */
+    private $submissionManager;
+
+    /**
+     * @return LoggerInterface
+     */
+    public function getLogger()
```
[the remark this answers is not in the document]
paveldhq: we discussed such possibility only

inc/Smartling/Helpers/ContentHelper.php
```diff
@@ -339,4 +340,26 @@ public function removeTargetMetadata(SubmissionEntity $submission)
...
+
+        try {
+            if (($this->getIoFactory()->getMapper($contentType)->get($contentId)) instanceof EntityAbstract) {
+                $result = true;
+            }
+        } catch (Exception $e) {
+        }
+
+        if ($needSiteSwitch) {
+            $this->getSiteHelper()->restoreBlogId();
```
dimitrystd: Can we add restore blog to `finally` block. Before it was not possible because of legacy php. But now we use 5.6 and it seems `finally` is supported.

vsolovei-smartling: A finally block would be redundant here

PR #216, opened 2019-09-16
implemented GlobalSettingsManager

paveldhq:


inc/Smartling/Services/GlobalSettingsManager.php
```diff
@@ -0,0 +1,215 @@
...
+    {
+        return SimpleStorageHelper::get(static::SELF_CHECK_IDENTIFIER, static::getSkipSelfCheckDefault());
+    }
+
+    public static function setSkipSelfCheck($value)
+    {
+        SimpleStorageHelper::set(static::SELF_CHECK_IDENTIFIER, $value);
+
+        if (static::getSkipSelfCheckDefault() === (int)$value) {
+            SimpleStorageHelper::drop(static::SELF_CHECK_IDENTIFIER);
```
PavelLoparev: Why do you need to drop  setting if it equals to default value? Here and below

paveldhq: generally yes,
1. Not to keep useless waste in database
2. In case that customer uses *default* value just keep using new default value without any db migrations.

PR #219, opened 2020-01-16
Missing change since "removed JSON.parse() calls because of using wp_send_json() on backend" (eb683411cb9520b6f16c13a6ebc32ace643f6817) commit

PavelLoparev:
The issue is not in WP `wp_send_json` function, it works fine. the issue is that we switched from manual response sending to `wp_send_json` this one which sets `Content-type: application/json` header and `jquery.post` method is smart enough to parse the response for you. So, the issue that we tried to parse already parsed string (js object).

We have a commit `removed JSON.parse() calls because of using wp_send_json() on backend` which removes all such `JSON.parse` line but one line wasn't removed.

Also, @dimitrystd, it's strange that at wp site our plugin has `1.11.1` version (https://wordpress.org/plugins/smartling-connector/) but in repo we have only `1.11.0` tag. Is something messed up with releases?

### DIFF
--- a/js/smartling-connector-admin.js
+++ b/js/smartling-connector-admin.js
@@ -296,7 +296,6 @@ jQuery(document).ready(function () {
                         submissionIds: submissionIds.join(",")
                     },
                     function (data) {
-                        data = JSON.parse(data);
                         switch (data.status) {
                             case "SUCCESS":
                                 wp.data.dispatch("core/notices").createSuccessNotice("Translations downloaded.");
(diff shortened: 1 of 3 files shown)


dimitrystd:
In svn 1.11.1. It seems we updated 3rd party libs only in svn
* https://plugins.svn.wordpress.org/smartling-connector/trunk/smartling-connector.php
* https://plugins.svn.wordpress.org/smartling-connector/trunk/readme.txt

PR #222, opened 2020-03-17
fix single item new job creation

vsolovei-smartling:


vsolovei-smartling:
> Did you check:
> 
> 1. Post edit page
> 2. Bulk submit page
> 3. Both pages with enabled/disabled manual dependency resolving?
> 
> I'm worrying about removed `locales = [$("#targetBlogId").val()];`

I've checked post edit page with automatic and manual relations handling (both with and without "only" checkbox), and the bulk submit handling.
As for the targetBlogId, the select was introduced in commit 07ae5c891289c600c1ca46b69c1da8da4895ca1f, and removed in the following 8d495b38b63bf2152a0204d13a52922b97732b51

PR #223, opened 2020-03-23
release v1.12.3

vsolovei-smartling:


PavelLoparev:
In general, seems like client left `MULTILINGUAL_PRESS_PRO_SITE_OPTION` option enabled in connector config

PR #227, opened 2020-05-12
add show exception on cron ttl not expired (WP-508)

vsolovei-smartling:


inc/Smartling/Jobs/JobAbstract.php
```diff
@@ -293,6 +294,8 @@ private function tryRunJob()
             if ($allowedToRun) {
                 $this->run();
                 $this->dropLockFlag();
+            } elseif ($message !== null) {
+                throw new FrontendSafeException($message);
```
PavelLoparev, line 298: Are you sure `tryRunJob()` is called only from one place (where you catch the `FrontendSafeException` exception)?

PavelLoparev: I mean there are at least 3 places in code where I see `do_action(DownloadTranslationJob::JOB_HOOK_NAME)` invocation

vsolovei-smartling: It's also run from
https://github.com/Smartling/wordpress-localization-plugin/blob/9dc18b75d351f65ab3ea3cddeb76311c0e5637ae/inc/Smartling/WP/Controller/ConfigurationProfilesController.php#L136
(both cron and manual link)

vsolovei-smartling: Done

inc/Smartling/Jobs/DownloadTranslationJob.php
```diff
@@ -15,6 +15,8 @@ class DownloadTranslationJob extends JobAbstract
 {
     const JOB_HOOK_NAME = 'smartling-download-task';
 
+    const SOURCE_FRONTEND = 'frontend';
```
dimitrystd, line 18: I would recommend `user`.

js/smartling-connector-admin.js
```diff
@@ -301,7 +301,11 @@ jQuery(document).ready(function () {
                                 wp.data.dispatch("core/notices").createSuccessNotice("Translations downloaded.");
                                 break;
                             case "FAIL":
-                                wp.data.dispatch("core/notices").createErrorNotice("Translations download failed.");
+                                let message = "Translations download failed.";
+                                if (data.message) {
+                                    message += "\n" + data.message;
```
dimitrystd, line 306: PHP_EOL?

readme.txt
```diff
@@ -58,6 +58,10 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
+= 1.12.7 =
+* Fixed issue where downloading content in Smartling widget erroneously reports that translations were downloaded even if they weren't
```
dimitrystd, line 62:
```
Fixed issue when Smartling widget erroneously reported that translations were downloaded even if they weren't
```

readme.txt
```diff
@@ -58,6 +58,10 @@ Additional information on the Smartling Connector for WordPress can be found [he
...
+* Moved older changelog entries to changelog.txt
```
dimitrystd, line 63: Not sure by i think
```
Moved old changelog entries to changelog.txt
```

inc/Smartling/WP/Controller/PostBasedWidgetControllerStd.php
```diff
@@ -130,10 +131,13 @@ public function ajaxDownloadHandler()
...
             $result = [];
             try {
-                do_action(DownloadTranslationJob::JOB_HOOK_NAME);
-                $result['status'] = 'SUCCESS';
+                do_action(DownloadTranslationJob::JOB_HOOK_NAME, DownloadTranslationJob::SOURCE_FRONTEND);
+                $result['status'] = self::RESPONSE_AJAX_STATUS_SUCCESS;
             } catch (\Exception $e) {
-                $result['status'] = 'FAIL';
+                $result['status'] = self::RESPONSE_AJAX_STATUS_FAIL;
+                if ($e instanceof FrontendSafeException) {
```
dimitrystd, line 138: If i run job manually why not to show all error messages? If it will be network timeout, or invalid DB password then it's still useful information. I understand that message will be less user-friendly but it's still useful.
And most important that we will get rid of this exception class with the weird name.

inc/Smartling/Exception/FrontendSafeException.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Smartling\Exception;
+
+class FrontendSafeException extends SmartlingExceptionAbstract
```
dimitrystd, line 5: Its name hurts my eyes. Left comment about it below

js/smartling-connector-admin.js
vsolovei-smartling: No, we're in javascript context, PHP_EOL might actually cause unwanted formatting

PR #230, opened 2020-05-27
"remove fields name" regex values support

vsolovei-smartling:


inc/Smartling/Helpers/FieldsFilterHelper.php
```diff
@@ -362,30 +361,31 @@ private static function getFieldProcessingParams()
...
-                $this->getLogger()->debug($debugMessage);
-                continue;
-            } else {
-                $rebuild[$key] = $value;
+        foreach ($fields as $key => $value) {
+            foreach ($removeRegexList as $regex) {
+                if (0 !== preg_match("/\/$regex$/", $key)) {
+                    $debugMessage = vsprintf('Removed field by name \'%s\' because of configuration.', [$key]);
+                    $this->getLogger()->debug($debugMessage);
+                    continue 2;
```
dimitrystd, line 381: Looks a little bit strange. If it's common for PHP then we can leave it as is. Otherwise i can propose instead of `$result` build a set of excluded fields. And then just `array_diff($removedFields, $fields)`

tests/Smartling/Helpers/FieldsFilterHelperTest.php
```diff
@@ -112,4 +112,38 @@ public function structurizeArrayDataProvider()
...
+            [
+                'meta/stays' => 'value',
+                'meta/tool_templates_1000_category' => 'value',
+                'meta/tool_templates_0_stays' => 'value',
+            ],
+            $x->removeFields($fields, ['tool_templates_\d{1,3}_(id|category|thumbnail_url|section)', 'irrelevantRegex']),
+            'Should remove fields that match regex list, fields that don\'t match should remain'
+        );
+
+        $fields = ['meta/stays' => 'value'];
```
dimitrystd, line 140: Let's keep tests as simple as possible. The current one contains at least 3 different cases.
Starting from this line is another test (with name like `testDonotRemoveFieldsIfEmptyRegex`).

I'm not sure what's the test convention we use in WP but in java i expected to see tests like this
- testFilterFieldsByRegularName
- testFilterFieldsByRegex - potentially added 2-3 tests with different regex + 1 regex in the list and 3 regexes
- testDonotRemoveFieldsIfEmptyRegex

inc/Smartling/Helpers/FieldsFilterHelper.php
```diff
@@ -362,30 +361,31 @@ private static function getFieldProcessingParams()
...
-        foreach ($array as $key => $value) {
-            if (1 === preg_match($pattern, $key)) {
-                $debugMessage = vsprintf('Removed field by name \'%s\' because of configuration.', [$key]);
-                $this->getLogger()->debug($debugMessage);
-                continue;
-            } else {
-                $rebuild[$key] = $value;
+        foreach ($fields as $key => $value) {
+            foreach ($removeRegexList as $regex) {
+                if (0 !== preg_match("/\/$regex$/", $key)) {
```
dimitrystd, line 378: Question about the second `$`. Does it mean that you always add end of line to any user regex? If so then it's bad idea. let's make flexible since we already allowed regex.

vsolovei-smartling: Done

tests/Smartling/Helpers/FieldsFilterHelperTest.php
vsolovei-smartling: Done

inc/Smartling/Helpers/FieldsFilterHelper.php
vsolovei-smartling: Would need to use array_diff_key here, I prefer to leave it as it is

PR #231, opened 2020-07-06
add display of last error for failed submissions, fix Gutenberg block attribute types when downloading translations

vsolovei-smartling:


inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -381,4 +382,38 @@ public function loadExternalDependencies()
...
+                                    break;
+                                case "number":
+                                    $attrs[$key] = (int)$attrs[$key];
+                                    break;
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
PavelLoparev, line 415: Tests are needed for all these branches

inc/Smartling/WP/WPAbstract.php
```diff
@@ -357,7 +358,21 @@ public static function localeSelectionCheckboxBlock($namePrefix, $blog_id, $blog
...
+     * @param int $percentage
+     * @param string $lastError
+     * @param array $statusFlags
+     * @return string
+     */
+    public static function localeSelectionTranslationStatusBlock(
+        $statusText,
+        $statusColor,
+        $percentage,
+        $lastError = '',
```
PavelLoparev, line 373: I would add this new parameter with the default value to the end just to make sure new signature won't break any calls

inc/Smartling/Helpers/GutenbergBlockHelper.php
vsolovei-smartling: Done

inc/Smartling/WP/WPAbstract.php
vsolovei-smartling: Done

PR #232, opened 2020-07-13
Use attribute types from original post

vsolovei-smartling:


inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -384,36 +383,19 @@ public function loadExternalDependencies()
...
+     * @param array $translatedAttributes
+     * @param array $originalAttributes
      * @return array
      */
-    private function fixAttributeTypes($name, array $attrs)
+    private function fixAttributeTypes($blockName, array $originalAttributes, array $translatedAttributes)
     {
-        if ($name === "core/image") { // For some reason, images are not in block type registry
-            $attrs["id"] = (int)$attrs["id"];
-        } elseif (class_exists(\WP_Block_Type_Registry::class)) {
```
PavelLoparev, line 395: `// For some reason, images are not in block type registry` did find out the reason for this? I mean this code branch is deleted in new code

vsolovei-smartling: There's no need to check registry, if the original block works, the translated one will work in the same way

PR #233, opened 2020-07-14
add locale when saving submission from upload widget (WP-522)

vsolovei-smartling:


inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -415,6 +415,7 @@ public function createSubmissionsHandler($data = '')
                 $submissionTemplateArray = [
                     SubmissionEntity::FIELD_SOURCE_BLOG_ID => $this->getContentHelper()->getSiteHelper()->getCurrentBlogId(),
                     SubmissionEntity::FIELD_TARGET_BLOG_ID => (int)$targetBlogId,
+                    SubmissionEntity::FIELD_TARGET_LOCALE => $this->localizationPluginProxy->getBlogLocaleById((int)$targetBlogId),
```
dimitrystd, line 418: Target blog id always has 1:1 relationship to target locale. I would say "target_locale" column is redundant, only for human. I don't understand why we need to add it into search query.

dimitrystd:
Close it without merge because of obsolete

PR #234, opened 2020-07-14
rework download translations button (WP-519)

vsolovei-smartling:


js/smartling-connector-admin.js
```diff
@@ -1,3 +1,8 @@
+/**
+ * @var ajaxurl
+ * @var wp
+ */
+var btnSelector = ".smtPostWidget-submitBlock input[value=Download]";
```
PavelLoparev, line 5: It appears that download button doesn't have id. Why not just add it and use it in selector instead of `value`?

image

js/smartling-connector-admin.js
```diff
@@ -7,6 +12,7 @@
             form: '#smartling-form',
             post_widget: '#smartling-post-widget',
             submit: '#submit',
+            download: '#smartling_download',
```
PavelLoparev, line 15: Nitpicking: mix of `-` and `_` here.

js/smartling-connector-admin.js
```diff
@@ -80,28 +85,9 @@
...
-            return false;
-        },
-
-        onTranslationSend: function (e) {
-            e.preventDefault();
-            var
-                data = $(this.selectors.post_widget).serializeArray();
-
-            console.log(e);
+            return this.validateFields($(this.selectors.form).serializeArray());
```
PavelLoparev, line 90: :+1:

js/smartling-connector-admin.js
```diff
@@ -80,28 +85,9 @@
...
-
-            return false;
-        },
-
-        onTranslationSend: function (e) {
-            e.preventDefault();
-            var
-                data = $(this.selectors.post_widget).serializeArray();
-
-            console.log(e);
```
PavelLoparev, line 104: Is it leftover of an old code which is not needed anymore?

js/smartling-connector-admin.js
PavelLoparev: Also, I don't see where you add this id to the button

js/smartling-connector-admin.js
```diff
@@ -254,65 +241,98 @@ jQuery(document).ready(function () {
...
+    dismissButton.setAttribute("type", "button");
+    dismissButton.setAttribute("title", "dismiss");
+    dismissButton.classList.add("notice-dismiss");
+    div.appendChild(dismissButton);
+    var h1 = document.getElementsByTagName("h1")[0];
+    h1.parentNode.insertBefore(div, h1.nextSibling);
+    dismissButton.addEventListener("click", function () {
+        div.parentNode.removeChild(div);
+    });
+}
```
PavelLoparev, line 323: Why don't you use jquery for all these manipulations with DOM?

js/smartling-connector-admin.js
```diff
@@ -203,18 +189,20 @@
     };
 
     $(function () {
-        var content = $('#smartling-form');
-        if (content.length > 0) {
+        if ($(localizationOptions.selectors.form).length > 0) {
             localizationOptions.init();
         }
+        if ($(localizationOptions.selectors.post_widget).length > 0) {
+            $(localizationOptions.selectors.download).on("click", function () {
```
PavelLoparev, line 196: I don't get how it works if there is no element by this selector

PavelLoparev:
Frankly speaking, I didn't get what is the key change. There are some refactoring in this PR. Is it `submit` vs `button`? How does it help?

js/smartling-connector-admin.js
vsolovei-smartling: Yes, no usages whatsoever, and this function is useless

js/smartling-connector-admin.js
vsolovei-smartling: There is an element. If none were present, that could would do nothing.

vsolovei-smartling:
> Frankly speaking, I didn't get what is the key change. There are some refactoring in this PR. Is it `submit` vs `button`? How does it help?

The button vs submit prevents post saving instead of downloading. An additional issue is that in this case, the button did nothing, so I've extracted a function and it get's used where previously the download wasn't handled properly (lines 195...)

js/smartling-connector-admin.js
vsolovei-smartling: Done

js/smartling-connector-admin.js
vsolovei-smartling: Fixed

PR #236, opened 2020-07-16
Fix some strings being cloned instead of sent to translation

vsolovei-smartling:


inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -22,19 +22,17 @@ class GutenbergBlockHelper extends SubstringProcessorHelperAbstract
...
     {
         $copyList = [
-            'type',
-            'providerNameSlug',
-            'align',
-            'className',
+            '^type$',
+            '^providerNameSlug$',
+            '^align$',
+            '^className$',
```
PavelLoparev, line 28: This is because we use regexp for copying?

vsolovei-smartling: This erroneously causes fields like section_0_type_testimonial to be cloned instead of being sent to translation

PR #241, opened 2020-08-13
mark submissions failed on unsuitable batch

vsolovei-smartling:


inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -523,14 +524,27 @@ public function bulkSubmit(SubmissionEntity $submission)
...
+                    ]);
+                    foreach ($submissions as $found) {
+                        $found->setStatus(SubmissionEntity::SUBMISSION_STATUS_FAILED);
+                    }
+                    $this->getSubmissionManager()->storeSubmissions($submissions);
+                    break;
+                }
+                $e = $e->getPrevious();
+            } while ($e !== null);
+            $e = $caught;
```
PavelLoparev, line 547: 1. It looks like fix for symptoms. The real issue is somewhere else, because connector somewhy tries to upload file into the executed batch.
2. I would suggest to research exact case of recent  `Batch is not suitable for adding files` issue and find out is it really connector's issue or it were duplicated queue items on prod/stg of one client or something like this.

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -523,14 +524,28 @@ public function bulkSubmit(SubmissionEntity $submission)
...
-            };
-            $proceedAuthException($e);
+                if ($submissionHasBatchUid
+                    && strpos("Batch is not suitable for adding files", $e->getMessage()) !== false) {
+                    $submissions = $this->getSubmissionManager()->find([
+                        SubmissionEntity::FIELD_STATUS => [SubmissionEntity::SUBMISSION_STATUS_NEW],
+                        SubmissionEntity::FIELD_BATCH_UID => [$submission->getBatchUid()],
+                    ]);
+                    foreach ($submissions as $found) {
+                        $found->setBatchUid('');
```
PavelLoparev, line 540: Do we have other places where we need to reset batch uid?

inc/Smartling/Base/SmartlingCoreUploadTrait.php
vsolovei-smartling: I've did some additional digging into it, and it seems some client have old submissions stuck with invalid batch uids

inc/Smartling/Base/SmartlingCoreUploadTrait.php
vsolovei-smartling: This looks like the only one, so I think this should solve the issue.

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -523,14 +524,28 @@ public function bulkSubmit(SubmissionEntity $submission)
...
+                        $found->setStatus(SubmissionEntity::SUBMISSION_STATUS_FAILED);
+                    }
+                    $this->getSubmissionManager()->storeSubmissions($submissions);
```
dimitrystd, line 543: Add here logging that you failed submissions. So we will have a chance to figure out why hundreds-thousands submissions will be failed.

readme.txt
```diff
@@ -58,6 +58,9 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
+= 1.13.6 =
+* Fixed daily bucket job uploads handling failure due to content being added to an unsuitable translation job
```
dimitrystd, line 62: No, this is not a fix. A fix should get rid of this error. In your case, you silently skip uploads.

PR #244, opened 2020-09-02
Remove failing submissions

vsolovei-smartling:


.phpstorm.meta.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace PHPSTORM_META {
+
+    use Smartling\Helpers\ArrayHelper;
+
+    override(ArrayHelper::first(0), elementType(0));
+}
```
PavelLoparev, line 8: What it used for?

inc/Smartling/ApiWrapper.php
```diff
@@ -895,4 +894,26 @@ public function setNotificationRecord(ConfigurationProfileEntity $profile, $spac
...
+    /**
+     * @param \Exception $e
+     */
+    public function isUnrecoverable(\Exception $e) {
+        $unrecoverableKeys = [
+            'file.not.found',
+        ];
+
+        $message = $e->getMessage();
+        if (strpos($message, 'Array') === 0) {
```
PavelLoparev, line 907: Why not just `return strpos($message, "file.not.found") !== FALSE`?

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -523,7 +523,6 @@ public function bulkSubmit(SubmissionEntity $submission)
             }
 
             $this->executeBatch($submission->getBatchUid(), $submission->getSourceBlogId());
-            $this->closeBatch($submission->getBatchUid());
```
PavelLoparev, line 526: Didn't get why we don't "close" batch now.

inc/Smartling/Jobs/LastModifiedCheckJob.php
```diff
@@ -149,15 +137,24 @@ protected function filterSubmissions(array $lastModifiedResponse, array $submiss
...
+     * @return SubmissionEntity[]
      */
     protected function processFileUriSet(array $submissions)
     {
         if (ArrayHelper::notEmpty($submissions)) {
             $submission = ArrayHelper::first($submissions);
-            $lastModified = $this->getApiWrapper()->lastModified($submission);
+            try {
+                $lastModified = $this->apiWrapper->lastModified($submission);
+            } catch (SmartlingNetworkException $e) {
```
PavelLoparev, line 150: For me `SmartlingNetworkException` is something about request/response timeouts, socket errors etc. But not valid 400 response from API.

I believe this was implemented year(s) ago, I just noticed this. No action points from my side

PavelLoparev (changes requested):

inc/Smartling/Base/SmartlingCoreUploadTrait.php
vsolovei-smartling: After analysing logs, I've found this causes more issues then it solves.

.phpstorm.meta.php
vsolovei-smartling: This is a newish feature of phpstorm, and will not get included into the svn repository (https://www.jetbrains.com/help/phpstorm/ide-advanced-metadata.html).
So far, it explains to phpstorm that ArrayHelper::first() returns an element from the array argument, removing the need to provide that data in comments. Everytime we wrote code like
```
/**
 * @var SubmissionEntity $submission
*/
$submission = ArrayHelper::first($submissions);
```
With this file, the completion and other benefits of typed variable is available without the comment

inc/Smartling/ApiWrapper.php
vsolovei-smartling: Simplified for now

inc/Smartling/Jobs/LastModifiedCheckJob.php
vsolovei-smartling: ApiWrapper::LastModified wraps everything into SmartlingNetworkException (https://github.com/Smartling/wordpress-localization-plugin/blob/cb8e211f016eaf0d8fbd49d3cdeda409a9c08c5c/inc/Smartling/ApiWrapper.php#L445)

.phpstorm.meta.php
PavelLoparev: I'm ok with this since `will not get included into the svn repository`

inc/Smartling/Base/SmartlingCoreUploadTrait.php
vsolovei-smartling: https://bt.smartling.net/browse/WP-538

PavelLoparev (approved):

PR #245, opened 2020-09-04
add workaround for submissions not being sent for translation (WP-533)

vsolovei-smartling:


inc/Smartling/Helpers/ContentHelper.php
```diff
@@ -242,27 +242,31 @@ public function readTargetMetadata(SubmissionEntity $submission)
...
+                $this->getLogger()->warning('Wrapper ' . get_class($wrapper) . " was unable to get post id $result after setting");
+            }
         }
 
-        $this->ensureRestoredBlogId();
+        try {
+            $this->ensureRestoredBlogId();
+        } catch (\LogicException $e) {
+            $this->getLogger()->debug('Suppressed exception: ' . $e->getMessage());
+        }
```
PavelLoparev, line 269: Should we re-throw exception here?

inc/Smartling/Base/SmartlingCoreTrait.php
```diff
@@ -123,7 +123,12 @@ protected function prepareTargetEntity(SubmissionEntity $submission, $forceClone
          * @var EntityAbstract $targetContent
          */
         $targetContent = $this->getContentHelper()->writeTargetContent($submission, $targetContent);
-        $submission->setTargetId($targetContent->getPK());
+        if (is_int($targetContent)) {
```
PavelLoparev, line 126: What was changed so `readSourceContent` started returning int and objects? Or it was always doing that?

PavelLoparev (commented):
In general, I didn't get the difference between manual and automatic job run and how this change fixes the issue

PR #248, opened 2020-09-16
Skip attachments with empty ids

PavelLoparev:


inc/Smartling/Helpers/RelativeLinkedAttachmentCoreHelper.php
```diff
@@ -134,17 +134,22 @@ protected function processString(&$stringValue)
...
-                                    $this->getParams()->getSubmission()->getTargetBlogId(),
-                                    (int)$attachmentId,
-                                    $this->getParams()->getSubmission()->getBatchUid()
-                                );
-                                $replacer->addReplacementPair($attachmentId, $attachment->getTargetId());
-                            } catch (SmartlingManualRelationsHandlingSubmissionCreationForbiddenException $e) {
-                                $this->getLogger()->notice("Skipping attachment id $attachmentId: was not uploaded " .
-                                    'due to manual relations handling');
+
+                            if (!empty($attachmentId)) {
```
dimitrystd, line 138: I think we must confirm that:
- It's not empty
- It's `int`, not a string or anything else

inc/Smartling/Helpers/RelativeLinkedAttachmentCoreHelper.php
```diff
@@ -134,17 +134,22 @@ protected function processString(&$stringValue)
...
+                                        (int)$attachmentId,
+                                        $this->getParams()->getSubmission()->getBatchUid()
+                                    );
+                                    $replacer->addReplacementPair($attachmentId, $attachment->getTargetId());
+                                } catch (SmartlingManualRelationsHandlingSubmissionCreationForbiddenException $e) {
+                                    $this->getLogger()->notice("Skipping attachment id $attachmentId: was not uploaded " .
+                                      'due to manual relations handling');
+                                }
+                            } else {
+                                $this->getLogger()->warning("Can not send attachment as it has empty id acfFieldId=${value}");
```
dimitrystd, line 152: And since it can be strings value then include into message field name and actual field value that we rejected

readme.txt
```diff
@@ -58,6 +58,9 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
+= 1.13.10 =
+* Fixed download issue if acf attachment has empty id inside Guttenberg block
```
dimitrystd, line 62:
```
Fixed download flow if ACF attachment field has an empty id inside Guttenberg block
```

tests/Smartling/Helpers/RelativeLinkedAttachmentCoreHelperTest.php
```diff
@@ -63,6 +63,59 @@ public function testProcessGutenbergBlockAcf($string, array $definitions, $sourc
...
+            $acf
+        ])->setMethods(null)->getMock();
+
+        $source = [$string];
+        $meta = [];
+        $content = $this->getMock(PostEntityStd::class);
+
+        $x->processor(new AfterDeserializeContentEventParameters($source, $submission, $content, $meta));
+
+        self::assertEquals(
```
dimitrystd, line 113: I didn't get this assert. You compare what you set in the line [107](https://github.com/Smartling/wordpress-localization-plugin/pull/248/files?diff=unified&w=1#diff-2749f9a3f7536bc537697298c9021757R107)

dimitrystd (changes requested):

tests/Smartling/Helpers/RelativeLinkedAttachmentCoreHelperTest.php
PavelLoparev: `$x->processor()` accepts first parameter by reference. I was surprised as well and also didn't get the first test in this file from the first time.

Actually, we can just remove this assertion. The only thing we assert in this test is that exception is bever thrown.

readme.txt
PavelLoparev: Done

inc/Smartling/Helpers/RelativeLinkedAttachmentCoreHelper.php
PavelLoparev: I'm not sure it can be strings value.

> include into message field name and actual field value that we rejected
everything else can be found in logs by `acfFieldId`

inc/Smartling/Helpers/RelativeLinkedAttachmentCoreHelper.php
PavelLoparev: added check for `is_numeric`

dimitrystd (approved):

PR #250, opened 2020-09-24
Fix target submission metadata to support flexible content from ACF

vsolovei-smartling:


inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -327,6 +335,7 @@ public function applyXML(SubmissionEntity $submission, $xml)
                     $this->getContentHelper()->removeTargetMetadata($submission);
                 }
                 $metaFields = self::arrayMergeIfKeyNotExists($lockedData['meta'], $metaFields);
+                $metaFields = $this->restoreAcfMetaFields($metaFields, $sourceFields['meta']);
```
dimitrystd, line 338: I think the line above must copy ALL meta fields, including transient.
It seems you implemented a workaround: if transient field was translated then you restore its value from the source. But it's not right approach. We should prevent modification of transient and other non-translatable fields by our connector.

inc/Smartling/Helpers/XmlEncoder.php
```diff
@@ -179,21 +180,43 @@ private static function prepareXPath($xmlString)
      * @param string           $content
      * @param SubmissionEntity $submission
      *
-     * @return array
+     * @return DecodedXml
      */
     public static function xmlDecode($content, SubmissionEntity $submission)
```
dimitrystd, line 185: I don't know why "decode" was in "encoder" class before, but it was a single method before. Now we have 3 methods related to decode. Let's reorginize code: move these methods to a new class or rename the current one.

dimitrystd (changes requested):

inc/Smartling/Helpers/XmlEncoder.php
vsolovei-smartling: Renamed class

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -820,4 +829,34 @@ public function createForTranslation($contentType, $sourceBlog, $sourceEntity, $
 
         return $this->getSubmissionManager()->storeEntity($submission);
     }
+
+    private function restoreAcfMetaFields(array $metaFields, array $sourceMetaFields)
+    {
+        if (count($this->acfCopyRules) === 0) {
+            $acfDynamicSupport = new AcfDynamicSupport($this->getSubmissionManager()->getEntityHelper());
+            $acfDynamicSupport->run();
+            $this->acfCopyRules = $acfDynamicSupport->getCopyRules();
```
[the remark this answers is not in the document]
vsolovei-smartling: It could get called multiple times if downloading in bulk

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -820,4 +829,34 @@ public function createForTranslation($contentType, $sourceBlog, $sourceEntity, $
 
         return $this->getSubmissionManager()->storeEntity($submission);
     }
+
+    private function restoreAcfMetaFields(array $metaFields, array $sourceMetaFields)
```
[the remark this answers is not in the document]
vsolovei-smartling: Moved

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -248,15 +251,20 @@ private static function arrayMergeIfKeyNotExists($lockedData, $translation)
...
             $this->prepareFieldProcessorValues($submission);
-            if ('' === $xml) {
-                $translation = [];
-            } else {
-                $translation = XmlEncoder::xmlDecode($xml, $submission);
+            if ('' !== $xml) {
+                $decoded = XmlEncoder::xmlDecode($xml, $submission);
+                $translation = $decoded->getFields();
+                $sourceFields = $decoded->getSourceFields();
+                if (!array_key_exists('meta', $sourceFields)) {
```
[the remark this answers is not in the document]
vsolovei-smartling: Done

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -248,15 +251,20 @@ private static function arrayMergeIfKeyNotExists($lockedData, $translation)
     public function applyXML(SubmissionEntity $submission, $xml)
     {
         $messages = [];
-        try {
+        $sourceFields = ['meta' => []];
+        $translation = [];
```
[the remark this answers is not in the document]
vsolovei-smartling: Done

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -70,7 +70,9 @@ private function readSourceContentWithMetadataAsArray(SubmissionEntity $submissi
             'meta'   => $this->getContentHelper()->readSourceMetadata($submission),
         ];
 
-        $source['meta'] = $source['meta'] ? : [];
+        if (!is_array($source['meta'])) {
```
dimitrystd, line 73: It looks very similar to what was before. But with new code if `$source['meta']` was string\int before then you just replace the value with an empty array. It looks ok, just wanted to draw your attention.

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -243,20 +243,26 @@ private static function arrayMergeIfKeyNotExists($lockedData, $translation)
...
 
+        try {
             $lockedData = $this->readLockedTranslationFieldsBySubmission($submission);
 
             $this->prepareFieldProcessorValues($submission);
-            if ('' === $xml) {
-                $translation = [];
-            } else {
-                $translation = XmlEncoder::xmlDecode($xml, $submission);
+            $xmlFields = [];
```
dimitrystd, line 257: This name is not clear when you see its usage 50 lines below. Let's call it more specific like `$sorcePostMetaFields` (your name is welcome). And initialize it with `meta` of the original post. If someone will need other source fields then I prefer to push him to update the code and fix tests rather than just add usage of this variable in one more place.

inc/Smartling/Helpers/FieldsFilterHelper.php
```diff
@@ -210,6 +211,9 @@ public function processStringsBeforeEncoding(SubmissionEntity $submission, array
         ContentSerializationHelper::prepareFieldProcessorValues($settingsManager, $submission);
         $profile = $settingsManager->getSingleSettingsProfile($submission->getSourceBlogId());
         $data = $this->prepareSourceData($data);
+        if ($strategy === self::FILTER_STRATEGY_UPLOAD && !GlobalSettingsManager::isAcfDisabled()) {
```
dimitrystd, line 214: Just want to confirm that we don't have other hooks before calling this method. I prefer to keep ACF logic aside from core fields handling. We may anticipate more such plugins and it will be not sexy to put all of their code here.

inc/Smartling/Helpers/XmlHelper.php
```diff
@@ -179,21 +175,43 @@ private static function prepareXPath($xmlString)
      * @param string           $content
      * @param SubmissionEntity $submission
      *
-     * @return array
+     * @return DecodedXml
      */
-    public static function xmlDecode($content, SubmissionEntity $submission)
+    public function xmlDecode($content, SubmissionEntity $submission)
```
dimitrystd, line 180: Did you remove `static` for testing?

readme.txt
```diff
@@ -58,11 +58,14 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
+= 1.14.0 =
+* New behaviour of Auto synchronize properties on translated page with source expert profile setting. When turned on, extends support of ACF flexible fields in metadata.
```
dimitrystd, line 62: Added more details. Mostly for us when we will review what and when was added.
```
* Major change in `Auto synchronize properties` behavior. Starting from this release connector always replace all metadata in a translated post with metadata from the source post. Then connector applies translations to freshly copied metadata.
* If you use ACF flexible fields then you should enable `Auto synchronize properties`. It will force connector always replicate ACF fields from source to target and keep flexible fields consistent
```

dimitrystd (commented):
It looks good. I still worry what we plan to release but it's better then before. In general I don't like ACF smell in core classes.

PavelLoparev (changes requested):
Have no additional comments. Let's fix what @dimitrystd mentioned and release it

inc/Smartling/Helpers/FieldsFilterHelper.php
vsolovei-smartling: This is the beginning of work with content

inc/Smartling/Helpers/XmlHelper.php
vsolovei-smartling: Yes

readme.txt
vsolovei-smartling: Done

inc/Smartling/Base/SmartlingCoreUploadTrait.php
vsolovei-smartling: It seems to me that the original intention was to work with arrays, since everywhere we check for this property we try to init it with empty array if it doesn't exist or is falsy, and then work with it as an array

inc/Smartling/Base/SmartlingCoreUploadTrait.php
vsolovei-smartling: Done

dimitrystd (approved):

PR #251, opened 2020-09-28
add-to-job-error-handling

vsolovei-smartling:


dimitrystd (approved):
So it seems it never worked as expected.

inc/Smartling/WP/View/ContentEditJob.php
```diff
@@ -655,9 +656,14 @@ function (data) {
...
                             default:
-                                uiShowMessage("FAILED", "Failed adding content to upload queue.");
+                                uiShowMessage("FAILED", message);
                                 break;
                         }
+                    }).fail(function (e) {
+                        if (e.responseJSON && e.responseJSON.response && e.responseJSON.response.message) {
+                            message = e.responseJSON.response.message;
+                        }
+                        $('#error-messages').html(message);
```
PavelLoparev, line 666: Maybe it would be better to use `append`? I mean there might be other errors in `#error-messages` container and your error will just override all of them.

vsolovei-smartling: The container gets cleared upon button click, that should not happen

PavelLoparev: Is `#error-messages` only connector related? I mean if it's from WP core template then we just wipe out all errors except our

vsolovei-smartling: Yes, it is added in the connector plugin widget

PR #252, opened 2020-10-02
Add setting to link taxonomies instead of translating when possible

vsolovei-smartling:


inc/Smartling/ContentTypes/CustomPostType.php
```diff
@@ -154,4 +152,61 @@ public function registerFilters()
...
+        $original = $contentRelationsDiscoveryService->findOriginalSubmission($element->taxonomy, $submission->getSourceBlogId(), $element->term_id);
+        if ($original === null) {
+            throw new EntityNotFoundException("Unable to find original submission for {$submission->getId()}");
+        }
+        $translated = $submissionManager->find([
+            SubmissionEntity::FIELD_CONTENT_TYPE => $element->taxonomy,
+            SubmissionEntity::FIELD_TARGET_BLOG_ID => $submission->getTargetBlogId(),
+            SubmissionEntity::FIELD_SOURCE_BLOG_ID => $original->getSourceBlogId(),
+            SubmissionEntity::FIELD_SOURCE_ID => $original->getSourceId(),
+        ]);
```
dimitrystd, line 202: Looks very similar to `ContentRelationsDiscoveryService.submissionExists`

inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -809,4 +806,47 @@ protected function normalizeReferences(array $references)
...
+    }
+
+    /**
+     * @param string $contentType
+     * @param int $targetBlogId
+     * @param int $targetId
+     * @param SubmissionEntity|null $previous
+     * @return SubmissionEntity|null
+     */
+    public function findOriginalSubmission($contentType, $targetBlogId, $targetId, SubmissionEntity $previous = null)
```
dimitrystd, line 836: I tried to understand this method but it's beyond my level.

Let's do it om example. I have a submission for taxonomy with
* sourceBlogId = 1
* sourceId = 1
* targetBlogId = 2
* targetId = 5

So I call function with params `("taxonomy", 2, 5)`. It finds my submission and calls itself with params `("taxonomy", 1, 1)`. And this is what I can understand. We have only one source blog and it should never appear in targetBlogId.
I'm afraid that your code can find something, but it can be not what you expected.

dimitrystd (changes requested):
Need to be discussed in voice

inc/Smartling/Bootstrap.php
```diff
@@ -213,7 +212,7 @@ public function testMultilingualPressPlugin()
         }
 
         if (false === $found) {
-            add_action('admin_init', function () {
+            add_action('admin_init', static function () {
```
PavelLoparev, line 215: What is the point for making callbacks static?

inc/Smartling/Bootstrap.php
```diff
@@ -222,7 +221,7 @@ public function testMultilingualPressPlugin()
     }
 
     public function updateGlobalExpertSettings () {
-        $data =& $_POST['params'];
+        $data = $_POST['params'];
```
PavelLoparev, line 224: I'm afraid somewhere in the code it's expected to see changes in `$_POST['params']` (by $data reference)

inc/Smartling/Bootstrap.php
```diff
@@ -447,10 +446,9 @@ private function initializeBuildInContentTypes(ContainerBuilder $di)
         ContentTypeNavigationMenuItem::register($di);
         ContentTypeNavigationMenu::register($di);
 
-        $handlers = [
-            'taxonomies' => (new Taxonomies($di)),
-            'posts'      => (new PostTypes($di)),
-        ];
+        // add supported taxonomies and post types for translation
+        new Taxonomies($di);
+        new PostTypes($di);
```
PavelLoparev, line 451: I don't get this. Something is happening in constructors?

inc/Smartling/ContentTypes/CustomTaxonomyType.php
```diff
@@ -62,16 +64,16 @@ public function registerWidgetHandler()
     {
         if ($this->getConfigParser()->hasWidget()) {
             $di = $this->getContainerBuilder();
-            $tag = 'wp.taxonomy.' . static::getSystemName();
+            $tag = 'wp.taxonomy.' . $this->getSystemName();
```
PavelLoparev, line 67: Seems like this never woked before since this method is not static

PavelLoparev (changes requested):

inc/Smartling/Bootstrap.php
vsolovei-smartling: Reader of the code may safely assume that the function changes no class properties

inc/Smartling/Bootstrap.php
vsolovei-smartling: No code changes $_POST['params'], I think it is safe to remove unneeded reference

inc/Smartling/Bootstrap.php
vsolovei-smartling: Yes, something is happening, the entire DI  I've tried to get rid of an unused variable and got unexpected results, so I've removed just the variable and added a comment above these

inc/Smartling/ContentTypes/CustomTaxonomyType.php
vsolovei-smartling: Php is forgiving in this case, it works

inc/Smartling/Bootstrap.php
PavelLoparev: But you can change object properties in static function by passing `$this` through `use`. I mean static keyword applied to callback does not guarantee it won't change something. Or maybe I'm missing something?

inc/Smartling/ContentTypes/CustomTaxonomyType.php
PavelLoparev: oh, it's not about method is static or not. it's about late static binding.

vsolovei-smartling:
Superseded by #255

PR #253, opened 2020-10-05
add blog switch logging

vsolovei-smartling:


inc/Smartling/Helpers/ContentHelper.php
```diff
@@ -117,36 +118,92 @@ public function ensureBlog($blogId)
         $this->setNeedBlogSwitch($this->getSiteHelper()->getCurrentBlogId() !== $blogId);
 
         if ($this->isNeedBlogSwitch()) {
+            $this->blogSwitches = [];
+            add_action('switch_blog', [$this, 'logSwitchBlog']);
```
PavelLoparev, line 122: Seems like this potentially adds action to hook multiple times per script execution. Maybe let's move this to constructor to not to remove this action every time in `ensureRestoredBlogId`?

inc/Smartling/Helpers/ContentHelper.php
```diff
@@ -117,36 +118,92 @@ public function ensureBlog($blogId)
...
     }
 
-    public function ensureSource(SubmissionEntity $submission)
+    public function logSwitchBlog() {
+        $backtrace = debug_backtrace();
+        foreach ($backtrace as $frame) {
+            if (array_key_exists('function', $frame) && array_key_exists('args', $frame)) {
+                $args = $frame['args'];
+                if ($frame['function'] === 'do_action' && $args[0] === 'switch_blog') {
+                    array_shift($args);
```
PavelLoparev, line 133: Didn't get why do you need to shift here. You don't need this var anywhere after this

inc/Smartling/Helpers/ContentHelper.php
```diff
@@ -117,36 +118,92 @@ public function ensureBlog($blogId)
...
+        return true;
+    }
+
+    /**
+     * @param array $frame backtrace frame
+     * @return string
+     */
+    private function getSwitchBlogString(array $frame) {
+        $template = "Switched to blog %d in %s:%d";
+        $blog = $frame['args'][1];
```
PavelLoparev, line 163: Let's check if `$frame['args'][1]` is set before accessing it

inc/Smartling/Helpers/ContentHelper.php
```diff
@@ -117,36 +118,92 @@ public function ensureBlog($blogId)
...
+    }
+
+    /**
+     * @return bool
+     */
+    private function validateSwitchBlog() {
+        $state = 'switch';
+        foreach ($this->blogSwitches as $frame) {
+            $this->getLogger()->debug($this->getSwitchBlogString($frame));
+            $context = $frame['args'][3];
```
PavelLoparev, line 147: Let's check if `$frame['args'][3]` is set before accessing it

PavelLoparev (changes requested):

inc/Smartling/Helpers/ContentHelper.php
vsolovei-smartling: Removed

PavelLoparev:
Seems like this approach only covers subsequential blog switches as errors, but seems like it's not enough. I propose to check stack traces for `switchBlogId`/`restoreBlogId` (connector's methods) and if there are no mentions of these functions then it means switch was caused by external plugin/theme. it's still possible to use connector's code and call switch/restore in wrong order though

inc/Smartling/Helpers/ContentHelper.php
vsolovei-smartling: Done

inc/Smartling/Helpers/ContentHelper.php
vsolovei-smartling: WordPress calculates hashes for hooks, no harm in adding multiple times

inc/Smartling/Helpers/ContentHelper.php
vsolovei-smartling: Done

PavelLoparev (approved):

PR #256, opened 2020-10-09
Ignore unregistered post types on related submissions search

vsolovei-smartling:


inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -619,132 +618,133 @@ public function actionHandler()
         $curBlogId     = $this->getContentHelper()->getSiteHelper()->getCurrentBlogId();
         $targetBlogIds = $this->getTargetBlogIds();
 
-        if ($this->getContentHelper()->checkEntityExists($curBlogId, $contentType, $id)) {
+        if (!$this->getContentHelper()->checkEntityExists($curBlogId, $contentType, $id)) {
```
dimitrystd, line 621: Before we returned the error and didn't execute code below. The new condition will write error to response but will continue to work.

dimitrystd (commented):
In general looks good. I left a comment which can be ok

inc/Smartling/Services/ContentRelationsDiscoveryService.php
PavelLoparev: Seems like lno because eventually it calls `wp_send_json` which calls `die` inside

inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -619,132 +618,133 @@ public function actionHandler()
...
+        $responseData = [
+            'originalReferences' => $detectedReferences,
+        ];
 
-            foreach ($targetBlogIds as $targetBlogId) {
-                foreach ($detectedReferences as $contentType => $ids) {
+        $registeredTypes = get_post_types();
+        foreach ($targetBlogIds as $targetBlogId) {
+            foreach ($detectedReferences as $contentType => $ids) {
+                if (in_array($contentType, $registeredTypes, true)) {
```
PavelLoparev, line 733: Let's log this case for troubleshooting purposes (if not in array)

PavelLoparev (changes requested):

PavelLoparev (approved):
Left comment about logging

inc/Smartling/Services/ContentRelationsDiscoveryService.php

PR #257, opened 2020-10-19
Ingest all blocks

vsolovei-smartling:
Previously, blocks got duplicated if multiple were used on same page. By adding the block index, we're able to process all blocks.


PavelLoparev (approved):
Seems ok to me

PR #259, opened 2020-10-26
Don't show taxonomy links menu on network admin

vsolovei-smartling:
Page not intended for use on network admin, requires specific source blog


PavelLoparev (approved):
This is because this page is blog related?

vsolovei-smartling:
> This is because this page is blog related?

Yes, a source blog is determined from the current blog. Not having this would require user to somehow select source blog for linking purposes

dimitrystd (approved):

PR #263, opened 2020-11-09
add wrappers for MultilingualPress

vsolovei-smartling:


inc/Smartling/DbAl/MultilingualPressConnector.php
```diff
@@ -248,18 +253,29 @@ public function unlinkObjects(SubmissionEntity $submission)
...
+        } catch (\Exception $e) {
+            $this->getLogger()->debug('Failed to unlink objects: ' . $e->getMessage());
+            return false;
+        }
     }
 
     private function isMultilingualPluginActive()
     {
-        return class_exists('Mlp_Helpers');
+        $expectedClasses = ['Mlp_Helpers', 'Mlp_Content_Relations', 'Mlp_Site_Relations'];
```
PavelLoparev, line 272: Does new code work with an old and new one? Do we need to support both versions of MP?

inc/Smartling/DbAl/MultilingualPressConnector.php
```diff
@@ -286,11 +301,6 @@ public function getBlogLanguageById($blogId)
      */
     public function getBlogNameByLocale($locale)
     {
-        /**
-         * @var \wpdb $wpdb
-         */
-        global $wpdb;
```
PavelLoparev, line 292: Maybe for some reason, we needed 2 conenctions? this and `$this->wpdb`?

inc/Smartling/DbAl/MultilingualPressConnector.php
vsolovei-smartling: We don't -- the plugin is obsolete for our purposes, we only use the supported of versions if available.

dimitrystd (approved):

inc/Smartling/DbAl/MultilingualPressConnector.php
vsolovei-smartling: No, they're both actually one connection

PavelLoparev (approved):

PR #265, opened 2020-11-13
WP-561-menu-items-translation

vsolovei-smartling:


dimitrystd (changes requested):
Let's invert logic. Instead of pass "flag" through pipeline we should change all callers to call pipeline (get related items) only if it's required.

js/smartling-connector-admin.js
```diff
@@ -284,14 +284,22 @@ function ajaxDownload() {
                         }
                         break;
                 }
-                if (wp && wp.data && wp.data.dispatch && wp.data.dispatch('core/notices').createSuccessNotice && wp.data.dispatch('core/notices').createErrorNotice) {
```
PavelLoparev, line 287: I remember it was added in previous PR. Why delete this now?

PavelLoparev (commented):
In general, seems like  you just added one condition into each content entity type upload flow.

Probably the question are all the existing tests passed? And do we have test for menu items specifically?

vsolovei-smartling:
> In general, seems like you just added one condition into each content entity type upload flow.
> 
> Probably the question are all the existing tests passed? And do we have test for menu items specifically?

Except the one uncoditional call, yes.
Test pass, I'm currently trying to implement a menu items test

js/smartling-connector-admin.js
vsolovei-smartling: The client in question reported it only makes matters worse for them

dimitrystd (approved):

dimitrystd (approved):

PR #267, opened 2020-12-30
add fix for quotes in json breaking the content

vsolovei-smartling:


inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -301,13 +302,19 @@ public function renderTranslatedBlockNode(\DOMElement $node)
...
+            });
+        }
         $attributes = 0 < count($attrs) ? ' ' . json_encode($attrs, JSON_UNESCAPED_UNICODE) : '';
         $content = implode('', $chunks);
         $result = ('' !== $content)
             ? vsprintf('<!-- wp:%s%s -->%s<!-- /wp:%s -->', [$name, $attributes, $content, $name])
             : vsprintf('<!-- wp:%s%s /-->', [$name, $attributes]);
 
-        if ((function_exists('acf_parse_save_blocks') && function_exists('acf_has_block_type')) && acf_has_block_type($name)) {
+        if ($isJson || (function_exists('acf_has_block_type') && acf_has_block_type($name))) {
```
PavelLoparev, line 317: Shouldn't be `&&`?

vsolovei-smartling: No, the second part adds slashes for acf blocks. Wordpress strips a single level of slashes while saving post content, because it expects content to be slashed before being sent for saving. If we would not add slashes here, the content would get truncated.

PR #269, opened 2021-01-28
Add media attachments UI (WP-568)

vsolovei-smartling:
Adds a new section in Fine Tune. Allows to specify paths to media attachment ids. The plugin will link the id to a translated asset id upon translation, if one exists.


inc/Smartling/DI.php
```diff
@@ -0,0 +1,88 @@
+<?php
+
+namespace Smartling;
+
+use Smartling\Exception\SmartlingConfigException;
+use Symfony\Component\Config\FileLocator;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;
+
+class DI
```
vsolovei-smartling, line 10: I initially tried to untangle DITrait, which is not in fact a trait, but the resulting changeset was too large. (https://bt.smartling.net/browse/WP-570)

inc/Smartling/Tuner/MediaAttachmentRulesManager.php
```diff
@@ -0,0 +1,52 @@
...
+        foreach (array_merge($this->getPreconfiguredRules(), $this->listItems()) as $rule) {
+            $result[] = new GutenbergReplacementRule($rule['block'], $rule['path']);
+        }
+
+        return $result;
+    }
+
+    public function register()
+    {
+    }
```
PavelLoparev, line 51: Is it empty by intention?

inc/Smartling/WP/Controller/AdminPage.php
```diff
@@ -51,12 +53,15 @@ private function processAction()
             case 'filter':
                 $manager = new FilterManager();
                 break;
+            case 'media':
+                $manager = (new DI())->fromContainer('media.attachment.rules.manager');
```
PavelLoparev, line 57: Shouldn't the `DI` be a singleton? You create new DI instance with own ContainerBuilder inside each time you call `new DI()`. It means you will have N containers with own service registry.

PavelLoparev: And in general seems like you've implemented `service locator` pattern which is treated as antipattern. You should inject  `'media.attachment.rules.manager'` dependency from outside rather than resolve it inside the code which depends on this service.

inc/Smartling/WP/Controller/MediaRuleForm.php
```diff
@@ -0,0 +1,68 @@
...
+        $manager->loadData();
+
+        if ('' === $id) {
+            $manager->add($data);
+        } else {
+            $manager->updateItem($id, $data);
+        }
+
+        $manager->saveData();
+        wp_redirect(get_site_url() . '/wp-admin/admin.php?page=smartling_customization_tuning');
```
PavelLoparev, line 66: Minor: maybe `smartling_customization_tuning` -> `smartling_media_attachments_settings`?

tests/Smartling/Helpers/RelativeLinkedAttachmentCoreHelperTest.php
```diff
@@ -8,16 +8,25 @@
...
+    private $mediaAttachmentRulesManager;
+
     protected function setUp()
     {
         WordpressFunctionsMockHelper::injectFunctionsMocks();
+        $this->mediaAttachmentRulesManager = $this->getMock(MediaAttachmentRulesManager::class);
+        $this->mediaAttachmentRulesManager->method('getGutenbergReplacementRules')->willReturn([
+            new GutenbergReplacementRule('image', 'id'),
+            new GutenbergReplacementRule('media-text', 'mediaId'),
+        ]);
```
PavelLoparev, line 29: Minor: why do you mock this return? Seems like it should return this array as default values since you have defined them in yml file as build in wordpress values

PavelLoparev (changes requested):
Left few minot comments and some related to new `DI` class.

Additionally, I would like to have a small demo (maybe you could show it on today's friday internal demo?) to see how it looks like.

inc/Smartling/Tuner/MediaAttachmentRulesManager.php
vsolovei-smartling: Yes, it is required by interface, but nothing to do for this particular manager

inc/Smartling/WP/Controller/MediaRuleForm.php
vsolovei-smartling: This is handled in the root controller, the url is correct

tests/Smartling/Helpers/RelativeLinkedAttachmentCoreHelperTest.php
vsolovei-smartling: I'm used to not working with di container in tests, it's faster to mock the object

inc/Smartling/WP/Controller/AdminPage.php
vsolovei-smartling: Fortunately, this could be reworked. I thought that the filter ui service is invoked from wordpress, and I'm unable to add any parameters

tests/Smartling/Helpers/RelativeLinkedAttachmentCoreHelperTest.php
PavelLoparev: There might be a situation when you change defaults in the YAML file but don't change the test and the test will pass, but in reality, something may be broken.

vsolovei-smartling: The results of the method's call are configuration dependent, yes, but here we are testing with two arbitrary rules and ensure that the translation goes as expected.

PavelLoparev: agree. resolved

PavelLoparev (approved):

inc/Smartling/Tuner/MediaAttachmentRulesManager.php
dimitrystd: Then why not remove `implements WPHookInterface`?

inc/Smartling/Services/SmartlingFilterUiService.php
```diff
@@ -3,14 +3,22 @@
...
 use Smartling\WP\Controller\AdminPage;
 use Smartling\WP\Controller\FilterForm;
+use Smartling\WP\Controller\MediaRuleForm;
 use Smartling\WP\Controller\ShortcodeForm;
 use Smartling\WP\WPHookInterface;
 
 class SmartlingFilterUiService implements WPHookInterface
 {
+    private $mediaAttachmentRulesManager;
+    public function __construct(MediaAttachmentRulesManager $mediaAttachmentRulesManager)
```
dimitrystd, line 17: I didn't get the reason to inject this manager from view. Other managers are instantiated right in this class.

dimitrystd (changes requested):

inc/Smartling/Services/SmartlingFilterUiService.php
vsolovei-smartling: It's not injected from view, but from DI container (https://github.com/Smartling/wordpress-localization-plugin/pull/269/files#diff-78c7355521f118c31dd783b39acaead8f82ff908569067475b749340eb8ed4feR29)

inc/Smartling/Tuner/MediaAttachmentRulesManager.php
vsolovei-smartling: Removed

PR #270, opened 2021-02-05
improve debug (WP-569)

vsolovei-smartling:


inc/Smartling/Helpers/ContentHelper.php
```diff
@@ -123,12 +123,16 @@ public function ensureBlog($blogId)
 
     public function logSwitchBlog() {
         $backtrace = debug_backtrace();
-        foreach ($backtrace as $frame) {
+        foreach ($backtrace as $index => $frame) {
             if (array_key_exists('function', $frame) && array_key_exists('args', $frame)) {
                 $args = $frame['args'];
-                if ($frame['function'] === 'do_action' && $args[0] === 'switch_blog' && $this->isConnectorSwitch($backtrace)) {
+                if ($frame['function'] === 'do_action' && $args[0] === 'switch_blog' && !$this->isConnectorSwitch($backtrace)) {
```
PavelLoparev, line 129: Not sure about this change. How did it work before?

PavelLoparev (changes requested):

inc/Smartling/Helpers/ContentHelper.php
vsolovei-smartling: That's the reason I've had to debug manually, it reported in-plugin change instead of the one I was looking for

dimitrystd (approved):
It's hard to understand how it works. Probably real test will give answers.

PR #273, opened 2021-03-19
add stripslashes when saving fine tuning forms

vsolovei-smartling:


dimitrystd (approved):
It's hard to tell from PR how actually will work. Hope it will not affect regex

PavelLoparev (commented):
Can we add one test that reproduces the issue?

vsolovei-smartling:
> Can we add one test that reproduces the issue?

I gave it some thought, but the issue is caused by wordpress' behaviour of adding slashes to GET/POST content (https://github.com/WordPress/WordPress/blob/master/wp-settings.php#L428). Adding a unit test that would check if a slash gets removed from content is impractical, because if that behaviour changes, we'll have a unit test that passes, but controllers will break again.

PR #278, opened 2021-04-28
fix search (WP-584)

vsolovei-smartling:


inc/Smartling/Helpers/QueryBuilder/Condition/ConditionBlock.php
```diff
@@ -45,12 +45,9 @@ public function addCondition(Condition $condition): void
         $this->conditions[] = $condition;
     }
 
-    /**
-     * @return Condition[]
-     */
-    public function getConditions(): array
+    public function isEmpty(): bool
```
dimitrystd, line 48: Sorry, but I didn't get the root cause (the new method is ok).

Was `conditions` null? Or did we forget to check `blocks` in `SubmissionManager.php`? I'm curious because we may miss root cause fix.

inc/Smartling/Helpers/QueryBuilder/Condition/ConditionBlock.php
```diff
@@ -45,12 +45,9 @@ public function addCondition(Condition $condition): void
...
     {
-        return $this->conditions;
+        return count($this->conditions) > 0 || count($this->blocks) > 0;
```
PavelLoparev, line 50: Didn't get why `empty` when conditions `> 0`.

PavelLoparev (commented):
I'm confused by the name of new method but rather than that looks ok to me

inc/Smartling/Helpers/QueryBuilder/Condition/ConditionBlock.php
vsolovei-smartling: Yes, blocks were not checked, causing a query without conditions, but with blocks having conditions to be replaced to empty condition

inc/Smartling/Helpers/QueryBuilder/Condition/ConditionBlock.php
vsolovei-smartling: Good catch, added a test

PavelLoparev (approved):

PR #280, opened 2021-05-13
add Gutenberg block relations processing

vsolovei-smartling:


inc/Smartling/Helpers/RelativeLinkedAttachmentCoreHelper.php
```diff
@@ -100,21 +91,13 @@ protected function processString(&$stringValue): void
...
-        if (0 < preg_match_all(self::CORE_GUTENBERG_BLOCK, $stringValue, $matches)) {
-            foreach ($matches[1] as $key => $_) {
-                $info = $this->processGutenbergBlock($matches[1][$key], $matches[2][$key]);
-                if ($info !== null) {
-                    $stringValue = str_replace("<!-- wp:core/{$matches[1][$key]} {$matches[2][$key]} -->", "<!-- wp:{$matches[1][$key]} {$info->getReplacement()} -->", $stringValue);
-                    $replacer->addReplacementPair(new ReplacementPair("wp-image-{$info->getSourceId()}", "wp-image-{$info->getTargetId()}"));
-                }
-            }
-        }
-
```
vsolovei-smartling, line 117: This was moved to GutenbergBlockHelper, and further upgraded

inc/Smartling/Replacers/ContentIdReplacer.php
```diff
@@ -0,0 +1,92 @@
...
+                    $this->contentType,
+                    $sourceBlogId,
+                    (int)$value,
+                    $targetBlogId,
+                    $submission->getJobInfoWithBatchUid()
+                )->getTargetId();
+
+                if ($dataType === 'string') {
+                    $value = (string)$targetId;
+                }
```
PavelLoparev, line 86: I don't get why in some cases we need to cast value to string

inc/Smartling/WP/View/AdminPage.php
```diff
@@ -53,7 +53,7 @@
         <?php $filterTable->display(); ?>
     </form>
 
-    <h3>Custom media attachment rules</h3>
+    <h3>Gutenberg block rules</h3>
```
PavelLoparev, line 56: So we dropped media attachments support or now we have some generic approach for all attachments?

PavelLoparev (commented):
I think demo is needed. Additionally, seems like tests are missing

PavelLoparev (commented):
I think demo is needed. Additionally, seems like tests are missing

PR #283, opened 2021-05-28
add workaround for download not available (WP-589)

vsolovei-smartling:


inc/Smartling/Base/SmartlingCoreDownloadTrait.php
```diff
@@ -92,6 +88,15 @@ public function downloadTranslationBySubmission(SubmissionEntity $entity)
...
         } catch (\Exception $e) {
+            if ($e instanceof SmartlingFileDownloadException) {
+                $xml = $this->getXML($entity);
+
+                if ($xml === '') {
+                    $this->getLogger()->info("Detected empty xml for submissionId={$entity->getId()}, applying");
+                    $this->applyXML($entity, $xml, $this->xmlHelper, $this->postContentHelper);
+                    return;
+                }
+            }
```
PavelLoparev, line 99: I don't get this error handling

vsolovei-smartling: Sometimes there is no content to translate, in this case, the xml would not get uploaded. When the plugin tries downloading related menu items, it encounters this exception.
The initial issue is fixed, because if the connector handles it, it changes related item ids (https://github.com/Smartling/wordpress-localization-plugin/pull/282)

PR #284, opened 2021-06-01
add inner Gutenberg blocks locking (WP-595)

vsolovei-smartling:


inc/Smartling/Helpers/PostContentHelper.php
```diff
@@ -2,57 +2,66 @@
...
+                $parts = explode('/', $lockedField);
+                $index = array_shift($parts);
+                if (!array_key_exists($index, $originalBlocks)) {
+                    $this->logger->notice("Unable to get content for locked Gutenberg block $lockedField, skipping lock");
+                    continue;
+                }
+                if (count($parts) > 0) {
+                    $translatedBlocks[$index] = $this->applyLockedField($parts, $originalBlocks[$index], $translatedBlocks[$index]);
+                } else {
+                    $translatedBlocks[$index] = $originalBlocks[$index];
```
PavelLoparev, line 39: I didn't get why we override translated block with source block content.

vsolovei-smartling: These are locked blocks, so we need to skip applying translations for them

PR #286, opened 2021-06-10
fix download issue on related content not found (WP-600)

vsolovei-smartling:


readme.txt
```diff
@@ -60,6 +60,9 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
+= 2.4.1 =
+* Fixed an issue with downloads
```
dimitrystd, line 64: Pls add detail about an issue. Unfortunately, I don't understand from code and there is no jira ticket.

dimitrystd (changes requested):

readme.txt
vsolovei-smartling: I've added a Jira ticket (https://bt.smartling.net/browse/WP-600) with more details. This issue occurs when RelativeLinkedAttachmentCoreHelper tries to get attachment ID for an ID, but that id is not an attachment, and getAttachmentId returns null.

dimitrystd: Then let it be
```
Fixed an issue with downloads when the post has related assets
```

dimitrystd (approved):

PavelLoparev (approved):

PR #287, opened 2021-06-13
workaround and logging for detected null id reference on widget upload

vsolovei-smartling:


dimitrystd (commented):
It's hard to tell is it a correct fix or no. There are no tests and I'm still not sure about root cause (where `$id` comes from).
Probably I will give snapshot to the client, but in general PR should be updated before merging to master.

inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -561,7 +561,10 @@ public function actionHandler(): void
             foreach ($detectedReferences as $contentType => $ids) {
                 if (in_array($contentType, $registeredTypes, true) || in_array($contentType, $taxonomies, true)) {
                     foreach ($ids as $id) {
-                        if (!$this->submissionManager->submissionExists($contentType, $curBlogId, $id, $targetBlogId)) {
+                        // TODO: find out when a null id gets added to the list, this should not happen
+                        if ($id === null) {
+                            $this->getLogger()->notice("Null id passed when processing detected references contentType=\"$contentType\"");
+                        } elseif (!$this->submissionManager->submissionExists($contentType, $curBlogId, $id, $targetBlogId)) {
```
PavelLoparev, line 567: Seems like after this fix we will never see this log record.

PavelLoparev (approved):

dimitrystd (commented):
The biggest problem that non of the existing tests are failed after code change and no new tests.

PR #289, opened 2021-07-01
add html escaping (WP-606)

vsolovei-smartling:


dimitrystd (approved):
It's hard to tell if new escaping will be ok. But I expect that all this modified code will render read-only markdown (not content inside input elements)

PavelLoparev (approved):

vsolovei-smartling:
> It's hard to tell if new escaping will be ok. But I expect that all this modified code will render read-only markdown (not content inside input elements)

Input elements are mostly rendered by HtmlTagGeneratorHelper::tag, that does escaping with htmlspecialchars

PR #292, opened 2021-07-06
fix escaping in nested Gutenberg block attributes (WP-598)

vsolovei-smartling:


PavelLoparev (commented):
LGTM but in ticket comments @dimitrystd mentioned completely different expected fix:
> Ideally, we should detect shortcodes and replace them with smartling placeholder.
Did you discuss the current fix?

readme.txt
```diff
@@ -60,6 +60,9 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
+= 2.4.6 =
+* Fix escaping in nested Gutenberg block attributes
```
dimitrystd, line 64:
```suggestion
* Fixed delivery of Gutenberg blocks where attributes have nested objects. Before appropriate escaping was done only for attributes on the top level only
```

PR #293, opened 2021-07-07
fix untitled submissions when sending from post widget (WP-608) 

vsolovei-smartling:


inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -630,4 +639,13 @@ private function storeWithJobInfo(SubmissionEntity $submission, JobEntityWithBat
...
         $this->submissionManager->storeEntity($submission);
     }
+
+    public function getPostTitle(int $id): string
+    {
+        $post = get_post($id);
+        if ($post instanceof \WP_Post) {
+            return $post->post_title;
+        }
+        return '';
```
PavelLoparev, line 649: Nitpicking, I would return NULL in this case.

PavelLoparev (approved):

inc/Smartling/Services/ContentRelationsDiscoveryService.php
dimitrystd: Before title was UNTITLED. What it be now for non posts?

PavelLoparev: I think it still will be UNTITLED since according to the code we set title only if title is not empty
```
$title = $this->getPostTitle($id);
if ($title !== '') {
    $submission->setSourceTitle($title);
}
```

vsolovei-smartling: Exactly, actual UNTITLED is set if field empty at https://github.com/Smartling/wordpress-localization-plugin/blob/2ff1be20bee047464e16336cd36020240ce8f453/inc/Smartling/Helpers/FileUriHelper.php#L23

PR #304, opened 2021-08-06
change links to translated content if possible (WP-612)

vsolovei-smartling:


inc/Smartling/Helpers/RelativeLinkedAttachmentCoreHelper.php
```diff
@@ -360,4 +370,29 @@ private function getReplacerForImgTag(string $tag, PairReplacerHelper $replacer)
...
+                $submission = ArrayHelper::first($this->submissionManager->find([
+                    SubmissionEntity::FIELD_TARGET_BLOG_ID => $targetBlogId,
+                    SubmissionEntity::FIELD_SOURCE_ID => $sourcePostId
+                ]));
+                if ($submission !== false) {
+                    $url = parse_url($this->wordpressProxy->get_blog_permalink($targetBlogId, $submission->getTargetId()), PHP_URL_PATH);
+                    if (is_string($url)) {
+                        $result->addReplacementPair(new ReplacementPair($href, $url));
+                    }
+                }
```
PavelLoparev, line 392: What happens if there is no submission in DB for some reason? I mean what is default behavior of `PairReplacerHelper` if we didn't call `$result->addReplacementPair(new ReplacementPair($href, $url));` on it?

PavelLoparev (approved):

inc/Smartling/Helpers/RelativeLinkedAttachmentCoreHelper.php
vsolovei-smartling: It iterates over all replacement pairs, if none are present, it would return original string

PR #305, opened 2021-08-12
fail submissions from locales not present in last modified response (WP-622)

vsolovei-smartling:


dimitrystd (changes requested):
It seems miscommunication.
If user unauthorized all strings for locale then we just don't update submission (keep its state New\In progress).
If user excluded all strings then locale must be present in response.

vsolovei-smartling:
I coded this before communicating with you, but decided to show the draft anyway.
It appears we will have to communicate that state "in progress" and 0% progress is the expected behaviour for this case. This could also happen due to connector misconfiguration.

PR #306, opened 2021-08-16
add force status check (WP-622)

vsolovei-smartling:
not sure if these changes are needed, the only scenario this will be useful is if someone has accidentally changed the submission data in database


dimitrystd (approved):
Would be nice to see a screenshot as part of PR. But in general, sounds like a good idea for possible troubleshooting

PR #307, opened 2021-08-16
Fix WordPress fatal error when plugin is activated before creating a network (WP-625)

vsolovei-smartling:
Fix in 2ad14c11085664948771898f5c84d11dcafb66c4, other two commits are refactoring and version bump

### DIFF
--- a/inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
+++ b/inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
@@ -430,7 +430,9 @@ class AcfDynamicSupport
                 } catch (SmartlingDirectRunRuntimeException $e) {
                     $dbDefinitions = [];
                     DiagnosticsHelper::addDiagnosticsMessage(
-                        'Failed to get ACF definitions from database. Please ensure that WordPress network is set up properly'
+                        'Failed to get ACF definitions from database.' .
+                        'Please ensure that WordPress network is set up properly.<br>' .
+                        "Exception message: {$e->getMessage()}"
                     );
                 }
 
--- a/inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
+++ b/inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
@@ -6,6 +6,7 @@ use Psr\Log\LoggerInterface;
 use Smartling\Base\ExportedAPI;
 use Smartling\Bootstrap;
 use Smartling\Exception\SmartlingConfigException;
+use Smartling\Exception\SmartlingDirectRunRuntimeException;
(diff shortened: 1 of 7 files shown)


inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
```diff
@@ -473,7 +425,14 @@ private function tryRegisterACF()
                 DiagnosticsHelper::addDiagnosticsMessage(implode('<br/>', $msg));
                 $this->getLogger()->notice('Automatic ACF support is disabled.');
             } else {
-                $dbDefinitions = $this->getDatabaseDefinitions();
+                try {
+                    $dbDefinitions = $this->getDatabaseDefinitions();
+                } catch (SmartlingDirectRunRuntimeException $e) {
+                    $dbDefinitions = [];
+                    DiagnosticsHelper::addDiagnosticsMessage(
+                        'Failed to get ACF definitions from database. Please ensure that WordPress network is set up properly'
```
PavelLoparev, line 433: Maybe let's print the exception message for more details?

readme.txt
```diff
@@ -60,13 +60,16 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
-= 2.4.14
+= 2.4.15 =
+* Fix fatal error when plugin is activated before creating a network
```
PavelLoparev, line 64: `Fixed` or `Fix for`

inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
dimitrystd: +1 to add the original $e message as the second line

dimitrystd (commented):
+1 for refactoring but I didn't get the fix. It seems just try-catch was added.

inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
vsolovei-smartling: Improved

readme.txt
vsolovei-smartling: Log entry removed, won't be in a release of it's own

vsolovei-smartling:
> +1 for refactoring but I didn't get the fix. It seems just try-catch was added.

This is the fix: without it the behaviour is to throw exception, doesn't get caught, user unable to use admin panels.

PavelLoparev (approved):

dimitrystd (approved):
It seems you have to pull master into the branch before merge it.

PR #308, opened 2021-08-17
add mlp3 support (WP-629)

vsolovei-smartling:


inc/Smartling/DbAl/MultilingualPress3Connector.php
```diff
@@ -0,0 +1,99 @@
...
+                $contentRelations = resolve(ContentRelations::class);
+                return $contentRelations->deleteRelation($this->getContentIds($submission), $submission->getContentType());
+            } catch (\Exception $e) {
+                return false;
+            }
+        }
+        return false;
+    }
+
+    public function getBlogNameByLocale(string $locale): string
```
dimitrystd, line 74: Implementation looks very similar to MLP2. Is it the same? Then maybe move to `MultilingualPressAbstract`

inc/Smartling/Helpers/LoggerSafeTrait.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Smartling\Helpers;
+
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
+use Smartling\MonologWrapper\MonologWrapper;
+
+trait LoggerSafeTrait {
```
dimitrystd, line 9: It looks like something reusable. But does this new logger support our configuration (what should or should not be logged)?

inc/Smartling/DbAl/MultilingualPress3Connector.php
vsolovei-smartling: Extracted

inc/Smartling/Helpers/LoggerSafeTrait.php
vsolovei-smartling: Yes. Trait created because many classes do the same call at some point in their lifecycle
smartling-connector/inc% grep -r 'MonologWrapper::getLogger(' * | wc -l
      40

inc/Smartling/Jobs/UploadJob.php
```diff
@@ -77,6 +82,7 @@ private function processUploadQueue(): void
             );
 
             do_action(ExportedAPI::ACTION_SMARTLING_SEND_FILE_FOR_TRANSLATION, $entity);
+            $this->localizationPlugin->linkObjects($entity);
```
PavelLoparev, line 85: Seems like instead of patching this class we could just react on this `ACTION_SMARTLING_SEND_FILE_FOR_TRANSLATION` hook somewhere and call `$this->localizationPlugin->linkObjects($entity);` there. I mean, in Drupal we usually create submodules to introduce support for some external modules by using hooks from the core module.

inc/Smartling/DbAl/MultilingualPress3Connector.php
```diff
@@ -0,0 +1,81 @@
...
+class MultilingualPress3Connector extends MultilingualPressConnector implements LocalizationPluginProxyInterface
+{
+    public function addHooks(): void
+    {
+    }
+
+    public function getBlogLocaleById(int $blogId): string
+    {
+        try {
+            return resolve(SiteSettingsRepository::class)->siteLanguageTag($blogId);
```
PavelLoparev, line 19: What does `resolve` do? Why can't we inject this `SiteSettingsRepository` dependency through DI?

vsolovei-smartling: It's external code, and resolve does just that: resolves an instance of the interface from their container

inc/Smartling/Jobs/UploadJob.php
vsolovei-smartling: Done

PavelLoparev (approved):

PR #312, opened 2021-08-23
visibility PR for fix fatal error at bootstrap

vsolovei-smartling:


readme-dev.md
```diff
@@ -1,4 +1,9 @@
 # Notes for developers
+## Building frontend assets
+```shell
+cd build && npm run build
```
dimitrystd, line 4: @ksoloviov-smartling There is the build step below (in docker). Do we need to change something there?

ksoloviov-smartling: @dimitrystd, the question is not for me, right?

PR #318, opened 2021-09-14
fix WP proxy (WP-636)

vsolovei-smartling:


inc/Smartling/Helpers/WordpressFunctionProxyHelper.php
```diff
@@ -64,7 +64,7 @@ public function get_terms()
 
     public function url_to_postid(string $url): int
     {
-        return (int)$this->url_to_postid($url);
+        return (int)url_to_postid($url);
```
dimitrystd, line 67: It looks like method calls itself. Infinite recursion?

tests/IntegrationTests/tests/AttachmentsTest.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace IntegrationTests\tests;
+
+use Smartling\Helpers\SiteHelper;
+use Smartling\Helpers\TranslationHelper;
+use Smartling\Submissions\SubmissionManager;
+use Smartling\Tests\IntegrationTests\SmartlingUnitTestCaseAbstract;
+
+class AttachmentsTest extends SmartlingUnitTestCaseAbstract
```
dimitrystd, line 10: Class name doesn't correlate with test name. Do we really rewrite external links as part of attachments handling?

dimitrystd (approved):

dimitrystd (changes requested):
Let's talk about how it may help client

inc/Smartling/Helpers/WordpressFunctionProxyHelper.php
PavelLoparev: I believe it was infinite recursion but now it's just wrapper (for tests purposes?) for `url_to_postid` wp function.

PavelLoparev (approved):

tests/IntegrationTests/tests/AttachmentsTest.php
vsolovei-smartling: We rewrite links to post based content

inc/Smartling/Helpers/WordpressFunctionProxyHelper.php
vsolovei-smartling: Yes

vsolovei-smartling:
> Let's talk about how it may help client

Without this, code ends up in infinite recursion. Some php configurations are able to catch this condition, most are not.

dimitrystd (approved):

PR #320, opened 2021-09-22
use distributed lock service instead of DB locking (WP-643)

vsolovei-smartling:


inc/Smartling/Jobs/JobAbstract.php
```diff
@@ -156,28 +69,33 @@ public function uninstall(): void
...
-    public function deactivate()
+    public function deactivate(): void
     {
         $this->uninstall();
     }
 
-    protected function getCronFlagName()
+    protected function getCronFlagName(): string
     {
         return 'smartling_cron_flag_' . $this->getJobHookName();
```
dimitrystd, line 84: Let's use `wordpress_smartling_cron_flag_` because we use global storage now

inc/Smartling/Jobs/JobAbstract.php
```diff
@@ -191,15 +109,24 @@ public function placeLockFlag($renew = false)
                 $flagName,
                 $this->getJobHookName(),
                 $newFlagValue,
-                $this->getWorkerTTL(),
+                $this->workerTTL,
             ]
         ));
 
-        SimpleStorageHelper::set($flagName, $newFlagValue);
+        if ($renew) {
```
dimitrystd, line 116: Let's remove `$renew` parameter because we can't use it by design

dimitrystd (changes requested):

inc/Smartling/Jobs/JobAbstract.php
```diff
@@ -156,28 +69,33 @@ public function uninstall(): void
...
-     * @param bool $renew
+     * @throws SmartlingApiException
      */
-    public function placeLockFlag($renew = false)
+    public function placeLockFlag(bool $renew = false): void
     {
+        $profile = ArrayHelper::first($this->settingsManager->getActiveProfiles());
+        if ($profile === false) {
+            $this->getLogger()->debug('No active profiles, skipping ' . $this->getJobHookName() . ' run');
+            return;
```
PavelLoparev, line 95: Don't we need to throw an exception here? What happens in client's code that calls `placeLockFlag` with no active profile? Will it still try to run some job?

inc/Smartling/Jobs/JobAbstract.php
```diff
@@ -191,15 +109,24 @@ public function placeLockFlag($renew = false)
...
     }
 
-    public function dropLockFlag()
+    public function dropLockFlag(): void
     {
+        $profile = ArrayHelper::first($this->settingsManager->getActiveProfiles());
+        if ($profile === false) {
+            $this->getLogger()->debug('No active profiles when trying to drop lock flag for ' . $this->getJobHookName());
+            return;
+        }
```
PavelLoparev, line 129: Same here. Also, seems like it's code duplication. Does it make sense to extract to somethink like `getActiveProfile()`?

inc/Smartling/Jobs/JobAbstract.php
```diff
@@ -210,87 +137,46 @@ public function dropLockFlag()
...
-                        [
-                            $this->getJobHookName(),
-                            $this->getWorkerTTL(),
-                            $flagTS - time(),
-                        ]
-                    );
-                    $this->getLogger()->debug($message);
-                }
-            }
+            $this->placeLockFlag();
```
PavelLoparev, line 165: So, yep, if there is no active profile you will try to still run this job and something will fail.

inc/Smartling/WP/Controller/PostBasedWidgetControllerStd.php
```diff
@@ -132,7 +132,7 @@ public function ajaxDownloadHandler()
             }
             $result = [];
             try {
-                do_action(DownloadTranslationJob::JOB_HOOK_NAME, DownloadTranslationJob::SOURCE_USER);
+                do_action(DownloadTranslationJob::JOB_HOOK_NAME);
```
PavelLoparev, line 135: Didn't get this change

PavelLoparev (changes requested):

inc/Smartling/Helpers/LoggerSafeTrait.php
```diff
@@ -7,7 +7,7 @@
 use Smartling\MonologWrapper\MonologWrapper;
 
 trait LoggerSafeTrait {
-    private LoggerInterface $logger;
```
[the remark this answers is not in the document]
vsolovei-smartling: Prevents "Typed property ... must not be accessed before initialization" errors

inc/config/cron.yml
```diff
@@ -3,7 +3,7 @@ parameters:
   cron.interval.group.submission: "3h"
   cron.interval.last-modified-check: "5m"
   cron.interval.download: "5m"
-  cron.job.default_ttl: 1200
+  cron.job.default_ttl: 60
```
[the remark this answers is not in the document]
vsolovei-smartling: With proper locking there is no need to hold the locks for 20 minutes.

inc/Smartling/WP/Controller/PostBasedWidgetControllerStd.php
vsolovei-smartling: With distributed locking, we no longer can provide the user with ttl left until expiry

inc/Smartling/Helpers/LoggerSafeTrait.php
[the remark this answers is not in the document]
PavelLoparev: Wow, that does sound like a bug in PHP's traits

inc/Smartling/Jobs/DownloadTranslationJob.php
```diff
@@ -2,36 +2,28 @@
...
+
+    public function __construct(
+        ApiWrapperInterface $api,
+        SettingsManager $settingsManager,
+        SubmissionManager $submissionManager,
+        string $jobRunInterval,
+        int $workerTTL,
+        QueueInterface $queue
+    ) {
+        parent::__construct($api, $settingsManager, $submissionManager, $jobRunInterval, $workerTTL);
```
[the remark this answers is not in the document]
PavelLoparev: No action points to current PR though

inc/Smartling/Jobs/JobAbstract.php
vsolovei-smartling: It is used appropriately

inc/config/cron.yml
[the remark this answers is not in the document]
vsolovei-smartling: Discussed with @dimitrystd, changed to 180. Should reduce it further after removing automatic relations handling.

inc/Smartling/Jobs/JobAbstract.php
vsolovei-smartling: Done

inc/Smartling/Jobs/JobAbstract.php
vsolovei-smartling: Thanks, good catch, will now throw exception

inc/Smartling/Jobs/DownloadTranslationJob.php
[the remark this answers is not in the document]
vsolovei-smartling: Sorry about that, my only defence is we never want/have time to refactor something. I will go with independent PRs approach in the future.

inc/Smartling/Jobs/JobAbstract.php
vsolovei-smartling: Done

inc/Smartling/Jobs/JobAbstract.php
vsolovei-smartling: Fixed

inc/Smartling/Jobs/JobAbstract.php
dimitrystd: Good point, we should interrupt job execution. It's similar to the case when we can't acquire the lock.

inc/Smartling/Jobs/JobAbstract.php
```diff
@@ -210,87 +137,46 @@ public function dropLockFlag()
...
+        } catch (SmartlingApiException $e) {
```
dimitrystd, line 166: I think we must catch all exceptions here because it's one of plugin "entry point". So ideal structure can be
```php
private function tryRunJob(): void
{
  try {
    $this->placeLockFlag();
    try {
      $this->run();
    // } catch (...) {...} You can add it only if you would like to handle specific exception of add more data into exception
    } finally {
      $this->dropLockFlag();
    }
  } catch (\Exception $exception) {
    // We log ALL exceptions here
  }
}
```

dimitrystd (changes requested):

inc/Smartling/WP/Controller/PostBasedWidgetControllerStd.php
vsolovei-smartling: I gave it a second thought, and a message saying we're unable to acquire lock is better than nothing, reverted

inc/Smartling/Jobs/JobAbstract.php
vsolovei-smartling: Done

readme.txt
```diff
@@ -62,6 +62,9 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
+= 2.7.0 =
+* Switched from database to external service locking for cron jobs
```
dimitrystd, line 66: Let's add details for the history. 
```
. It avoids the issue when hosting doesn't allow SQL `SELECT ... FOR UPDATE`
```

dimitrystd (approved):

dimitrystd (approved):

PavelLoparev (approved):

PR #322, opened 2021-09-30
change cloning workflow (WP-647)

vsolovei-smartling:
When cloning submissions, they previously applied XML immediatelyat upload job run. This could cause issues when a posts relation was not yet processed, resulting in copying the relation ids instead of altering them as needed. After this PR, every cloned submission first creates target content, then XML is applied with all existing relations having proper target ids.

### DIFF
--- a/inc/Smartling/Jobs/UploadJob.php
+++ b/inc/Smartling/Jobs/UploadJob.php
@@ -22,6 +22,8 @@ class UploadJob extends JobAbstract
 
         $this->processUploadQueue();
 
+        $this->processCloning();
+
         $this->processDailyBucketJob();
 
         $this->getLogger()->info('Finished UploadJob.');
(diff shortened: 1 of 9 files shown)


inc/Smartling/Jobs/UploadJob.php
```diff
@@ -58,6 +60,18 @@ private function processUploadQueue(): void
...
+    private function processCloning(): void
+    {
+        $submissions = $this->submissionManager->findSubmissionsForCloning();
+        foreach ($submissions as $submission) {
+            do_action(ExportedAPI::ACTION_SMARTLING_PREPARE_SUBMISSION_UPLOAD, $submission);
+        }
+        // needs to be in a separate loop to have all relations when cloning
+        foreach ($submissions as $submission) {
+            do_action(ExportedAPI::ACTION_SMARTLING_CLONE_CONTENT, $submission);
+        }
```
PavelLoparev, line 72: Do we really need to first process ALL submissions with `ACTION_SMARTLING_PREPARE_SUBMISSION_UPLOAD`? I wonder if we can have just one loop that calls `ACTION_SMARTLING_PREPARE_SUBMISSION_UPLOAD` -> `ACTION_SMARTLING_CLONE_CONTENT` in one iteration

PavelLoparev (commented):
Left some comments. I got the initial idea but I'm not sure will it break something or not (I'm surprised not so many tests are changed after such code change)

inc/Smartling/Jobs/UploadJob.php
vsolovei-smartling: Yes, that's the idea for this pull requests. First create all target content, then change the ids as necessary.

PavelLoparev: I mean, why not:
```
foreach ($submissions as $submission) {
            do_action(ExportedAPI::ACTION_SMARTLING_PREPARE_SUBMISSION_UPLOAD, $submission);
             do_action(ExportedAPI::ACTION_SMARTLING_CLONE_CONTENT, $submission);
        }
        ```

PR #324, opened 2021-10-11
scope dependenies (WP-648)

vsolovei-smartling:
Build process changed to produce dependencies with changed namespace

PR #325, opened 2021-10-12
fix post based widget (WP-652)

vsolovei-smartling:


PavelLoparev (commented):
Isn't it the same as https://github.com/Smartling/wordpress-localization-plugin/pull/324?

readme.txt
```diff
@@ -64,6 +64,7 @@ Additional information on the Smartling Connector for WordPress can be found [he
 == Changelog ==
 = 2.8.0 =
 * Scoped dependencies to prevent conflicts with other installed versions of GuzzleHttp and Symfony libraries
+* Fixed content not being added to upload from post widget
```
dimitrystd, line 67: Can you add details here? Or maybe example. I didn't get from code what exact case you fixed. And why there is no tests?

PR #328, opened 2021-10-20
add support for nested Gutenberg blocks (WP-565)

vsolovei-smartling:


tests/Smartling/Helpers/GutenbergBlockHelperTest.php
```diff
@@ -375,6 +373,19 @@ public function renderGutenbergBlockDataProvider(): array
                 [],
                 '<!-- wp:acf/image-text-wrap {\"data\":{\"content\":\"[caption id=\\\\\\\u0022attachment_3000029563\\\\\\\u0022]\\\\\\\n\\\\\\\nText\",\"_content\":\"More text\"}} /-->',
             ],
+            'root block' => [
+                'acf/offset-impact',
+                $blockForNestedTest,
+                [],
+                '<!-- wp:acf/offset-impact {\"id\":\"block_5fe115ebd752d\",\"name\":\"acf\\\\/offset-impact\",\"data\":{\"content\":\"<img class=\\\\\"alignnone size-medium wp-image-3000025130\\\\\" src=\\\\\"https:\\\\/\\\\/example.com\\\\/ideas\\\\/wp-content\\\\/uploads\\\\/sites\\\\/4\\\\/2020\\\\/12\\\\/GettyImages-1179050298_twitter-800x450.jpg\\\\\" alt=\\\\\"\\\\\" width=\\\\\"800\\\\\" height=\\\\\"450\\\\\" \\\\/>\",\"_content\":\"field_5d64117a9210b\",\"media\":\"3000025132\",\"_media\":\"field_5d64118f9210c\",\"add_video_url\":\"\",\"_add_video_url\":\"field_5eb3d3092a4ca\"},\"align\":\"\",\"mode\":\"auto\",\"wpClassName\":\"wp-block-acf-offset-impact\"} /-->',
```
dimitrystd, line 380: As for me, there is too many slashes `acf\\\\/offset-impact`, maybe php requires something besides json.

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -221,7 +221,7 @@ public function getPostContentBlocks(string $string): array
     }
 
     #[ArrayShape(['attributes' => 'array', 'chunks' => 'array'])]
-    public function sortChildNodesContent(\DOMNode $node, SubmissionEntity $submission): array
+    public function sortChildNodesContent(\DOMNode $node, SubmissionEntity $submission, int $level = 0): array
```
dimitrystd, line 224: I partially agree to have a default value in `renderTranslatedBlockNode`. But I think we should have a strong contract here and define as `... int $level): array`.
We know that this function is part of recursion then we must to take all measures to prevent common recursion issues:
- Try to avoid recursion by using stack. It requires significant development so probably skip this item for now
- Make signature clear how it will be used in recursion. No default values, the caller must set `level` explicitly
- Explicit condition that will terminate the recursion. Let's throw exception or just stop recursion if `level > 4`. Will see how crazy wp devs can be if they will hit this limit.

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -278,24 +278,24 @@ private function processTranslationAttributes(SubmissionEntity $submission, stri
...
 
-    public function renderGutenbergBlock(string $name, array $attrs = [], array $chunks = []): string
+    public function renderGutenbergBlock(string $name, array $attrs = [], array $chunks = [], int $level = 0): string
     {
-        $isJson = $this->isJson($attrs);
-        if ($isJson) {
+        $needsQuotes = $this->isJson($attrs);
+        if ($needsQuotes && $level === 0) {
             array_walk_recursive($attrs, static function (&$value) {
                 $value = str_replace('&quot;', '"', $value);
```
dimitrystd, line 300: Strange that we escape only quotes. Can you tell what parser requires this? (HTML; json). I prefer to use some library function rather than str_replace. The current usage may raise security concerns.

vsolovei-smartling: This is required for json parsing. It's not escaping, it's a workaround for user content that uses \u0022 to encode quotes, which then get replaced by Smartling to &quot;, hence only one value to replace. Without this, we would get \u0026quot; after working with json.

inc/Smartling/Helpers/GutenbergBlockHelper.php
vsolovei-smartling: Not having a default value is a good point, done. Not sure about limiting recursion though, I can easily see it exceeding "sane" levels in pretty basic page builders that would use Gutenberg blocks, if any exist

dimitrystd (approved):

PR #334, opened 2021-10-27
add deleted submission status (WP-659)

vsolovei-smartling:
WIP, delete status that matches other connectors


inc/Smartling/Submissions/SubmissionManager.php
```diff
@@ -177,36 +180,9 @@ public function searchByBatchUid(string $batchUid): array
...
     }
 
-    /**
-     * Gets SubmissionEntity from database by primary key
-     * alias to getEntities
-     *
-     * @return null|SubmissionEntity[]
-     */
-    public function getEntityById(int $id): ?array
+    public function getEntityById(int $id): ?SubmissionEntity
```
dimitrystd, line 183: To be honest, it confuses that request by **Id** will not return deleted submission. As for me, it's a single way to get raw data from DB. Or let's never return deleted and get rid of comment in the method above `Deleted submissions ARE NOT excluded`
Same story for `findByIds()`

inc/Smartling/Submissions/SubmissionManager.php
```diff
@@ -277,6 +247,7 @@ public function find(array $params = [], int $limit = 0): array
             $block->addCondition($condition);
         }
 
+        $block->addCondition(Condition::getCondition(ConditionBuilder::CONDITION_SIGN_NOT_EQ, SubmissionEntity::FIELD_STATUS, [SubmissionEntity::SUBMISSION_STATUS_DELETED]));
```
dimitrystd, line 250: You made a major change in behavior but non of unit-tests failed. Let's add unit-tests to lock the new behavior (never return DELETED submissions)

dimitrystd (changes requested):

vsolovei-smartling:
Stale

PR #338, opened 2021-11-03
change core image class id (WP-661)

vsolovei-smartling:


inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -283,6 +283,19 @@ private function processTranslationAttributes(SubmissionEntity $submission, stri
         return $this->fixAttributeTypes($originalAttributes, $processedAttributes);
     }
 
+    private function processTranslationChunks(string $blockName, array $originalAttributes, array $translatedAttributes, array $chunks): array
+    {
+        $result = [];
+        if ($blockName === 'core/image') {
+            foreach ($chunks as $chunk) {
+                $result[] = preg_replace("/<img(.+)class=\"wp-image-{$originalAttributes['id']}\"/", "<img\$1class=\"wp-image-{$translatedAttributes['id']}\"", $chunk);
```
PavelLoparev, line 291: Is it possible that there are more classes than just one `wp-image-<id>` in `class` property?

vsolovei-smartling: Good point, thanks, fixed

tests/IntegrationTests/tests/GutenbergBlocksTest.php
```diff
@@ -88,4 +88,49 @@ public function testInnerBlocks()
         }
         $this->assertEquals("\n<p>[Ñ~ót á ~Gúté~ñbé~rg bl~óck]</p>", $blocks[1]->getInnerHtml(), 'Expected non-Gutenberg block to be translated');
     }
+
+    public function testCoreImageClassTranslation()
```
dimitrystd, line 92: Great that you added the test, but it's an integration test (that includes a couple of services). I worry about long-term maintenance. As for me, such simple code can be tested with unit-test (test only `renderTranslatedBlockNode`)

dimitrystd (approved):

tests/IntegrationTests/tests/GutenbergBlocksTest.php
vsolovei-smartling: This would take more time to implement than doing an integration test, because one would need to dive deep into DOMElement and construct a proper one to test, created https://bt.smartling.net/browse/WP-665

PR #342, opened 2021-11-22
add locking retries (WP-667)

vsolovei-smartling:


inc/Smartling/Jobs/JobAbstract.php
```diff
@@ -209,6 +231,27 @@ private function getActiveProfile(): ConfigurationProfileEntity
         return $profile;
     }
 
+    private function lockWithRetry(callable $command, $ignoreLockNotAcquired = false): void
```
dimitrystd, line 234: It looks ok, we can release. But I expected something more sophisticated since we use symfony DI and can transparently do some magic around `api` object. I thought that we can create a "proxy" class (implements `ApiWrapperInterface`) but it intercepts all method calls by using PHP magic methods `__invoke`.

dimitrystd (approved):

inc/Smartling/Jobs/JobAbstract.php
vsolovei-smartling: https://bt.smartling.net/browse/WP-669, I'm strongly against using magic methods for this, needs discussion

PR #345, opened 2021-11-24
add test run frontend (WP-662)

vsolovei-smartling:
New section
image
Backend basis for test run included, will need to act on results in a future PR


inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -518,8 +514,14 @@ private function addPostContentReferences(array $array, GutenbergBlock $block):
...
 
-    public function actionHandler(): void
+    /**
+     * @return void|array
+     */
+    public function actionHandler(array $data = null)
     {
+        if ($data !== null) {
+            $this->requestData = $data;
+        }
```
PavelLoparev, line 524: So, you've already refactored `BaseAjaxServiceAbstract` constructor to receive request data. Why do you need to pass the same here once again? Why don't we pass needed parameters to `$_GET` from UI?

PavelLoparev (changes requested):

inc/Smartling/WP/Controller/TestRunController.php
```diff
@@ -0,0 +1,188 @@
...
+                'job' =>
+                    [
+                        'id' => $job->getJobUid(),
+                        'name' => $job->getJobName(),
+                        'description' => '',
+                        'dueDate' => '',
+                        'timeZone' => 'UTC',
+                        'authorize' => 'true',
+                    ],
+                'targetBlogIds' => $targetBlogId,
```
PavelLoparev, line 156: Shouldn't it be `[$targetBlogId]`?

PavelLoparev (changes requested):

inc/Smartling/Services/ContentRelationsDiscoveryService.php
PavelLoparev: Oh, because it's used only for logging and this method is also called from the loop with different params. Ok, but looks strange anyway

PavelLoparev: Actually no, it's not ok) The problem is that `ContentRelationsDiscoveryService` class mixes inside the controller and service layer. And `actionHandler` method is a controller that returns json response. And the problem is that you are trying to use this controller as a service to get relations. I think we need to separate the service layer, inject it to the `ContentRelationsDiscoveryService` and use it there. And leave this class (do not modify its returning value depending on `$data === null`). What you should do is that use that new service layer class in you `testRun` method instead of calling `ContentRelationsDiscoveryService::actionHandler`

inc/Smartling/WP/Controller/TestRunController.php
vsolovei-smartling: No, it's passed from frontend and expects a comma-delimited string

inc/Smartling/Services/ContentRelationsDiscoveryService.php

inc/Smartling/WP/Controller/TestRunController.php
PavelLoparev: got it

PR #351, opened 2021-12-13
use preg_quote to avoid specifying delimiters (WP-674)

vsolovei-smartling:


inc/Smartling/Tuner/MediaAttachmentRulesManager.php
```diff
@@ -34,12 +34,12 @@ public function getGutenbergReplacementRules(?string $blockType = null, ?string
         $rules = array_merge($this->preconfiguredRules, $this->listItems());
         if ($blockType !== null) {
             $rules = array_filter($rules, static function ($item) use ($blockType) {
-                return preg_match("#{$item->getBlockType()}#", $blockType) === 1;
+                return preg_match('#' . preg_replace('~([^\\\\])#~', '\1\#', $item->getBlockType()) . '#', $blockType) === 1;
```
dimitrystd, line 37: I didn't get what it does but thanks for the tests. Will sleep well because it seems to work

dimitrystd (approved):

inc/Smartling/Tuner/MediaAttachmentRulesManager.php
vsolovei-smartling: I tried using preg_quote function, but it also unnecessarily slashes special regex characters users might need. This slashes only `#`, and only if it was not slashed by the user themselves, in order not to introduce unneeded slashing

PR #353, opened 2021-12-16
fix media duplication when cloning (WP-675)

vsolovei-smartling:
While I was unable to reproduce the issue, I believe it boils down to unexpected content types when cloning

### DIFF
--- a/inc/Smartling/Helpers/ContentHelper.php
+++ b/inc/Smartling/Helpers/ContentHelper.php
@@ -65,10 +65,7 @@ class ContentHelper
         $this->ioFactory = $ioFactory;
     }
 
-    /**
-     * @return SiteHelper
-     */
-    public function getSiteHelper()
+    public function getSiteHelper(): SiteHelper
     {
         return $this->siteHelper;
     }
--- a/inc/Smartling/Services/ContentRelationsDiscoveryService.php
+++ b/inc/Smartling/Services/ContentRelationsDiscoveryService.php
@@ -181,9 +181,22 @@ class ContentRelationsDiscoveryService
                 }
             }
             foreach ($sources as $source) {
-                $submissionArray[SubmissionEntity::FIELD_CONTENT_TYPE] = $source['type'];
                 $submissionArray[SubmissionEntity::FIELD_SOURCE_ID] = (int)$source['id'];
-                $existing = ArrayHelper::first($this->submissionManager->find($submissionArray));
+                $found = $this->submissionManager->find($submissionArray);
+                if (count($found) > 1) {
+                    $types = array_reduce($found, static function (array $carry, SubmissionEntity $submission) {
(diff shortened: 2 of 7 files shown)


inc/Smartling/Helpers/ContentHelper.php
```diff
@@ -65,10 +65,7 @@ public function setIoFactory($ioFactory)
         $this->ioFactory = $ioFactory;
     }
 
-    /**
-     * @return SiteHelper
-     */
-    public function getSiteHelper()
+    public function getSiteHelper(): SiteHelper
```
vsolovei-smartling, line 68: Easier to write tests, phpunit mocks functions with known return types automatically

inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -181,9 +181,22 @@ public function clone(array $data): void
                 }
             }
             foreach ($sources as $source) {
-                $submissionArray[SubmissionEntity::FIELD_CONTENT_TYPE] = $source['type'];
```
vsolovei-smartling, line 184: We ignore type because there should be no multiple submissions with same source blog id and source id, but different types.

PavelLoparev: In what cases it's possible to have multiple submissions per source/target blog id? Should we stop cloning in such cases instead of making an assumption that the first submission is ok to proceed with?

tests/Services/ContentRelationDiscoveryServiceTest.php
```diff
@@ -354,9 +355,6 @@ public function testJobInfoGetsStoredOnNewSubmissions()
 
             $x = $this->getContentRelationDiscoveryService($apiWrapper, $contentHelper, $settingsManager, $submissionManager);
 
-            if (function_exists('switch_to_blog')) {
-                switch_to_blog(1);
-            }
```
PavelLoparev, line 359: What was it for?

vsolovei-smartling: I believe I've tried to improve this test, but the function is stubbed, and this call makes no difference

inc/Smartling/Services/ContentRelationsDiscoveryService.php
vsolovei-smartling: The relations discovery service could potentially in some cases incorrectly determine content type. We need to clone at least one content item, so I don't think stopping is ok in that case.

PavelLoparev: I'm worried that we try to fix something we can't reproduce and we make assumptions like `could potentially in some cases incorrectly determine content type`. In what exactly cases? This sounds like we need to fix the relations discovery service at first. because current fix just hides consequences.

PR #354, opened 2021-12-22
fix media library items duplication when cloning (WP-675)

vsolovei-smartling:
What previously happened was an attachment being created two times: first because it was detected by related media processor, second because it was queued for cloning. Now we get submissions for cloning not in bulk, but one by one, thus on second run we already know it was cloned to the target blog

### DIFF
--- a/inc/Smartling/Jobs/UploadJob.php
+++ b/inc/Smartling/Jobs/UploadJob.php
@@ -62,12 +62,8 @@ class UploadJob extends JobAbstract
 
     private function processCloning(): void
     {
-        $submissions = $this->submissionManager->findSubmissionsForCloning();
-        foreach ($submissions as $submission) {
+        while (($submission = $this->submissionManager->findSubmissionForCloning()) !== null) {
             do_action(ExportedAPI::ACTION_SMARTLING_PREPARE_SUBMISSION_UPLOAD, $submission);
-        }
-        // needs to be in a separate loop to have all relations when cloning
-        foreach ($submissions as $submission) {
             do_action(ExportedAPI::ACTION_SMARTLING_CLONE_CONTENT, $submission);
         }
     }
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
   "name": "smartling/wordpress-connector",
   "license": "GPL-2.0-or-later",
-  "version": "2.11.0",
+  "version": "2.11.1",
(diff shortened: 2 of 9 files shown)


inc/Smartling/Jobs/UploadJob.php
```diff
@@ -62,12 +62,8 @@ private function processUploadQueue(): void
...
     private function processCloning(): void
     {
-        $submissions = $this->submissionManager->findSubmissionsForCloning();
-        foreach ($submissions as $submission) {
+        while (($submission = $this->submissionManager->findSubmissionForCloning()) !== null) {
             do_action(ExportedAPI::ACTION_SMARTLING_PREPARE_SUBMISSION_UPLOAD, $submission);
-        }
-        // needs to be in a separate loop to have all relations when cloning
-        foreach ($submissions as $submission) {
             do_action(ExportedAPI::ACTION_SMARTLING_CLONE_CONTENT, $submission);
```
PavelLoparev, line 67: Do we change the submission "state" or "is cloned" flag so that it won't be returned during the next iteration?

vsolovei-smartling: Yes, the state gets changed to completed after cloning

PavelLoparev (approved):

inc/Smartling/Jobs/UploadJob.php
```diff
@@ -62,12 +62,8 @@ private function processUploadQueue(): void
 
     private function processCloning(): void
     {
-        $submissions = $this->submissionManager->findSubmissionsForCloning();
-        foreach ($submissions as $submission) {
+        while (($submission = $this->submissionManager->findSubmissionForCloning()) !== null) {
             do_action(ExportedAPI::ACTION_SMARTLING_PREPARE_SUBMISSION_UPLOAD, $submission);
-        }
-        // needs to be in a separate loop to have all relations when cloning
-        foreach ($submissions as $submission) {
```
PavelLoparev, line 70: I remember I commented something last time about these 2 loops. I do not remember why but you 100% needed exactly this order of hooks fired. Do we break that case by fixing this?

vsolovei-smartling: No, the case still stands, the prepare_upload action doesn't clone the contents

dimitrystd (commented):
+1

PR #359, opened 2022-01-26
fix sending if no related content present (WP-685)

vsolovei-smartling:
visibility PR, issue discovered while manually testing 2.11.2 after its release

### DIFF
--- a/inc/Smartling/Services/ContentRelationsHandler.php
+++ b/inc/Smartling/Services/ContentRelationsHandler.php
@@ -80,13 +80,16 @@ class ContentRelationsHandler extends BaseAjaxServiceAbstract
      *      'relations'    => {{@see actionHandler }} relations response
      *  ]
      */
-    public function createSubmissionsHandler(): void
+    public function createSubmissionsHandler(array $data = null): void
     {
+        if ($data === null) {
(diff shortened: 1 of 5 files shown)


PavelLoparev (approved):

inc/Smartling/Services/ContentRelationsHandler.php
```diff
@@ -80,13 +80,16 @@ public function register(): void
      *      'relations'    => {{@see actionHandler }} relations response
      *  ]
      */
-    public function createSubmissionsHandler(): void
+    public function createSubmissionsHandler(array $data = null): void
```
dimitrystd, line 83: Optional parameter triggered me.
As far as I understand this function is called from our JS function. Can we do this magic `$data = $_POST;` on browser side and always expect input data in the same format?

dimitrystd (changes requested):

inc/Smartling/Services/ContentRelationsHandler.php
vsolovei-smartling: No, we can't do that, php doesn't allow expressions as default values. The function will always be called without a parameter in production, this change was made for testing.

PR #366, opened 2022-03-10
improve Gutenberg block support (WP-690)

vsolovei-smartling:
Simplified and improved slashing, should now only add slashes for ACF, because ACF strips slashes from own blocks.
Fixed issues when a block's property was an array.

### DIFF
--- a/inc/Smartling/Helpers/GutenbergBlockHelper.php
+++ b/inc/Smartling/Helpers/GutenbergBlockHelper.php
@@ -102,7 +102,7 @@ class GutenbergBlockHelper extends SubstringProcessorHelperAbstract
         }
         $result = '';
         $originalBlocks = $this->parseBlocks($original);
-        $translatedBlocks = $this->parseBlocks(stripslashes($translated));
+        $translatedBlocks = $this->parseBlocks($translated);
         if (count($originalBlocks) !== count($translatedBlocks)) {
             $this->getLogger()->notice('Counts of blocks differ between original and translated, skipping replacing of post translate block content');
             return $translated;
@@ -110,7 +110,7 @@ class GutenbergBlockHelper extends SubstringProcessorHelperAbstract
         foreach ($translatedBlocks as $index => $block) {
             $result .= $this->wpProxy->serialize_block($this->applyPostTranslationReplacers($originalBlocks[$index], $block)->toArray());
         }
-        return addslashes($result);
+        return $result;
     }
 
     private function applyPostTranslationReplacers(GutenbergBlock $original, GutenbergBlock $translated): GutenbergBlock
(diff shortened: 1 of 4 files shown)


inc/Smartling/Helpers/RelativeLinkedAttachmentCoreHelper.php
```diff
@@ -318,7 +318,7 @@ private function getReplacerForAcfGutenbergBlock(string $block): PairReplacerHel
             $acfData = json_decode(stripslashes($block), true, 512, JSON_THROW_ON_ERROR);
             if (array_key_exists('data', $acfData)) {
                 foreach ($acfData['data'] as $key => $value) {
-                    if (array_key_exists($value, $this->acfDefinitions)
+                    if ((is_string($value) || is_int($value)) && array_key_exists($value, $this->acfDefinitions)
```
PavelLoparev, line 321: Looks unrelated. Also, seems like this condition is not covered by tests.

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -380,21 +380,20 @@ public function renderGutenbergBlock(string $name, array $attrs, array $chunks,
...
-        if ($needsQuotes && $depth === 0) {
+        To properly encode the quotes in JSON again we replace &quot; with \" */
+        if ($this->isPossibleJson($attrs) && $depth === 0) {
             array_walk_recursive($attrs, static function (&$value) {
-                $value = str_replace('&quot;', '"', $value);
+                $value = str_replace('&quot;', '\"', $value);
             });
         }
-        $attributes = 0 < count($attrs) ? ' ' . json_encode($attrs, JSON_UNESCAPED_UNICODE) : '';
+        $attributes = 0 < count($attrs) ? ' ' . json_encode($attrs, JSON_THROW_ON_ERROR | JSON_UNESCAPED_UNICODE) : '';
```
dimitrystd, line 389: We can expect new exceptions from the new code. Why don't we handle them?

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -380,21 +380,20 @@ public function renderGutenbergBlock(string $name, array $attrs, array $chunks,
...
         $content = implode('', $chunks);
         $result = ('' !== $content)
             ? vsprintf('<!-- wp:%s%s -->%s<!-- /wp:%s -->', [$name, $attributes, $content, $name])
             : vsprintf('<!-- wp:%s%s /-->', [$name, $attributes]);
 
-        // Only apply slashes for root depth: ACF blocks and blocks that we have previously replaced quotes in
-        if ($depth === 0 && ($needsQuotes || (function_exists('acf_has_block_type') && acf_has_block_type($name)))) {
+        // ACF will call stripslashes on save, but $result is properly slashed now
```
PavelLoparev, line 395: Is this true for all versions of ACF or for some specific one?

inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -380,21 +380,20 @@ public function renderGutenbergBlock(string $name, array $attrs, array $chunks,
...
+        if ($this->isAcfBlock($name)) {
```
PavelLoparev, line 396: Previously there was also a condition for `depth === 0`. We don't need it anymore here?

PavelLoparev (commented):
I think I need a call/chat to fully understand the whole picture of the issue.

inc/Smartling/Helpers/RelativeLinkedAttachmentCoreHelper.php
PavelLoparev: Oh, sorry, you've mentioned `Fixed issues when a block's property was an array.`.

PavelLoparev: Anyway, tests are needed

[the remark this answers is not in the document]
vsolovei-smartling: No, we could make it traverse array values, but this falls outside of the tasks' scope

inc/Smartling/Helpers/GutenbergBlockHelper.php
vsolovei-smartling: Reworked

inc/Smartling/Helpers/GutenbergBlockHelper.php
vsolovei-smartling: Previously I chose to add slash via addslashes call at the end, now it's been moved upwards, so the block only adds quotes to properly registered ACF blocks

inc/Smartling/Helpers/GutenbergBlockHelper.php
vsolovei-smartling: It's true since ACF 5.7.13 (three years ago)

dimitrystd (approved):

inc/Smartling/Helpers/GutenbergBlockHelper.php
PavelLoparev: Do we lock this specific dependency for connector some how?

inc/Smartling/Helpers/GutenbergBlockHelper.php
PavelLoparev: Did you push your commit? I don't see this change.

PR #367, opened 2022-03-16
fix translation request handling (WP-609)

vsolovei-smartling:
I've previously introduced a regression in handling, while fixing 2 levels deep cloning issues. Fixed, tests added

### DIFF
--- a/inc/Smartling/ContentTypes/AutoDiscover/PostTypes.php
+++ b/inc/Smartling/ContentTypes/AutoDiscover/PostTypes.php
@@ -2,6 +2,8 @@
 
 namespace Smartling\ContentTypes\AutoDiscover;
 
+use Smartling\Base\ExportedAPI;
+
 class PostTypes
 {
     private array $ignoredTypes;
@@ -17,7 +19,7 @@ class PostTypes
             return;
         }
 
-        add_action('smartling_register_custom_type', static function (array $definition) use ($postType) {
+        add_action(ExportedAPI::FILTER_SMARTLING_REGISTER_CUSTOM_POST_TYPE, static function (array $definition) use ($postType) {
             global $wp_post_types;
 
             if (true === $wp_post_types[$postType]->public) {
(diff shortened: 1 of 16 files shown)


inc/Smartling/ContentTypes/AutoDiscover/PostTypes.php
```diff
@@ -17,7 +19,7 @@ public function hookHandler($postType): void
             return;
         }
 
-        add_action('smartling_register_custom_type', static function (array $definition) use ($postType) {
+        add_action(ExportedAPI::FILTER_SMARTLING_REGISTER_CUSTOM_POST_TYPE, static function (array $definition) use ($postType) {
```
vsolovei-smartling, line 22: Not related to the WP-609 issue, but I don't really wan't to create a PR just for this single change

inc/Smartling/Models/CloneRequest.php
```diff
@@ -45,4 +43,19 @@ public function getTargetBlogIds(): array
     {
         return $this->targetBlogIds;
     }
+
+    public static function fromArray(array $array): self
+    {
+        return new self(self::getSourceId($array), $array['source']['contentType'], $array['relations'] ?? [], explode(',', $array['targetBlogIds']));
```
PavelLoparev, line 49: What if arry doesn't contain `['source']['contentType']`?

inc/Smartling/Models/TranslationRequest.php
```diff
@@ -0,0 +1,50 @@
...
+class TranslationRequest extends CloneRequest
+{
+    private JobInformation $jobInformation;
+    private array $ids;
+
+    public function __construct(int $contentId, string $contentType, array $relations, array $targetBlogIds, JobInformation $jobInformation, array $ids)
+    {
+        parent::__construct($contentId, $contentType, $relations, $targetBlogIds);
+        $this->jobInformation = $jobInformation;
+        $this->ids = self::toIntegerArray($ids);
```
PavelLoparev, line 16: Why do we need own `toIntegerArray` defined in this class? Why not dyrectly use `ArrayHelper::toArrayOfIntegers()` here?

vsolovei-smartling: To avoid writing the same error message twice

PavelLoparev (approved):

PR #368, opened 2022-03-18
add support for nested JSON (WP-694)

vsolovei-smartling:


inc/Smartling/Helpers/GutenbergBlockHelper.php
```diff
@@ -125,12 +126,48 @@ private function applyPostTranslationReplacers(GutenbergBlock $original, Gutenbe
                 } catch (\InvalidArgumentException $e) {
                     // do nothing, $value is preserved
                 }
+                $translated = $this->setValue($translated, $rule, $value);
```
PavelLoparev, line 129: `setValue` throws now. Do we need to handle exceptions here?

inc/Smartling/Models/GutenbergBlock.php
```diff
@@ -47,10 +55,18 @@ public function getInnerBlocks(): array
         return $this->innerBlocks;
     }
 
+    public function setAttributes(array $attributes): self
```
PavelLoparev, line 58: Probably should be `withAttributes`

inc/Smartling/Models/GutenbergBlock.php
```diff
@@ -39,6 +39,14 @@ public function getSmartlingLockId(): ?string
         return $this->attributes[PostContentHelper::SMARTLING_LOCK_ID] ?? null;
     }
 
+    public function clearInnerBlocks(): self
```
PavelLoparev, line 42: Seems like should be `withNoInnerBlocks`.

inc/Smartling/Tuner/MediaAttachmentRulesManager.php
```diff
@@ -39,6 +39,9 @@ public function getGutenbergReplacementRules(?string $blockType = null, ?string
         }
         if ($attribute !== null) {
             $rules = array_filter($rules, static function ($item) use ($attribute) {
+                if (self::isJsonPath($item->getPropertyPath())) {
+                    return explode('.', $item->getPropertyPath())[1] === $attribute;
```
PavelLoparev, line 43: Ok, you probably assume `($item->getPropertyPath()` will always have a valid json path string. But if it returns just `$.` then `explode('.', $item->getPropertyPath())[1]` will fails I believe. Let's check if `[1]` element exists after exploding.

PavelLoparev (changes requested):

inc/Smartling/Helpers/GutenbergBlockHelper.php
vsolovei-smartling: Added logging

inc/Smartling/Tuner/MediaAttachmentRulesManager.php
vsolovei-smartling: No, explode('.', '$.') returns an array with two elements '$' and '', I'll skip the check for now

inc/Smartling/Models/GutenbergBlock.php
```diff
@@ -47,10 +55,18 @@ public function getInnerBlocks(): array
...
+        $result->attributes = $attributes;
+
+        return $result;
+    }
+
     /**
      * @param mixed $value
      */
-    public function withAttribute(GutenbergBlock $block, string $attribute, $value): self
+    public function withAttribute(string $attribute, $value): self
```
vsolovei-smartling, line 69: GutenbergBlock $block was not used in the function body

PR #372, opened 2022-04-18
add audit logs (WP-700)

vsolovei-smartling:


inc/Smartling/ApiWrapperWithRetries.php
```diff
@@ -45,6 +45,13 @@ public function releaseLock(ConfigurationProfileEntity $profile, string $key): v
         });
     }
 
+    public function auditLogCreate(SubmissionEntity $submission, string $actionType, string $description, ?bool $isAuthorize = null): void
+    {
+        $this->withRetry(function () use ($submission, $actionType, $isAuthorize, $description) {
+            $this->base->auditLogCreate($submission, $actionType, $description, $isAuthorize);
+        });
+    }
```
PavelLoparev, line 53: Can this be implemented as a magic method call instead of copypasting this wrapper `withRetry` call for every new method in the class? No action points, just thinking

inc/Smartling/ApiWrapper.php
```diff
@@ -111,6 +113,27 @@ public function releaseLock(ConfigurationProfileEntity $profile, string $key): v
             ->releaseLock($key);
     }
 
+    public function auditLogCreate(SubmissionEntity $submission, string $actionType, string $description, ?bool $isAuthorize = null): void
+    {
+        $profile = $this->getConfigurationProfile($submission);
+        $record = new CreateRecordParameters();
```
PavelLoparev, line 119: Let's add some more data:

Client fields (should simplify user identification):
- setClientUserEmail
- setClientUserName

Env info fields:
- setEnvId

Client data fields:
- setClientData (should we log any wp related data to the event like submission id/submission created/etc)

PavelLoparev (changes requested):

PavelLoparev (commented):
lgtm. left some comments

inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -605,4 +608,10 @@ private function getSources(CloneRequest $request, int $targetBlogId): array
 
         return $sources;
     }
+
+    private function logSubmissionCreated(SubmissionEntity $submission, string $description, bool $isAuthorize): void
+    {
+        $this->getLogger()->info(sprintf("Adding submissionId=%d, fileName=%s, sourceBlogId=%d, sourceId=%d for upload to batchUid=%s due to user request, reason=\"%s\"", $submission->getId(), $submission->getStateFieldFileUri(), $submission->getSourceBlogId(), $submission->getSourceId(), $submission->getBatchUid(), $description));
+        $this->apiWrapper->auditLogCreate($submission, CreateRecordParameters::ACTION_TYPE_UPLOAD, $description, $isAuthorize);
+    }
```
PavelLoparev, line 616: This log states `due to user request` but is `storeWithJobInfo` always being called due to user request? I mean daily job for instance. I think this log should be moved outside this `logSubmissionCreated` method and put right before creating log record. And then there will be no sense in `logSubmissionCreated` at all

inc/Smartling/ApiWrapperWithRetries.php
PavelLoparev: Actually, I think this class can contain only one magic method and look like this:
```
class ApiWrapperWithRetries implements ApiWrapperInterface {
    ...

    public function __construct(ApiWrapperInterface $base)
    {
        $this->base = $base;
    }

    __call($name, $parameters) {
        $this->withRetry(function () use ($name, $parameters) {
            all_user_func_array(array($this->base, $methodName), $parameters);
        });
    }
}
```

vsolovei-smartling: It could, but we have methods that need no retrying, and it would still have to implement the methods in the interface. I know this looks strange, but we don't often modify ApiWrapper methods

inc/Smartling/ApiWrapper.php
vsolovei-smartling: @dimitrystd, we discussed GDPR requirements, do you know whom could I contact to clarify whether we could use the name and email?

inc/Smartling/Services/ContentRelationsDiscoveryService.php
vsolovei-smartling: At the moment, yes, the basis for the function call depends on a recently introduced TranslationRequest class, that is intended for frontend requests.

inc/Smartling/ApiWrapper.php
```diff
@@ -111,6 +113,27 @@ public function releaseLock(ConfigurationProfileEntity $profile, string $key): v
             ->releaseLock($key);
     }
 
+    public function auditLogCreate(SubmissionEntity $submission, string $actionType, string $description, ?bool $isAuthorize = null): void
```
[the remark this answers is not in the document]
dimitrystd: Agree, other methods also start with verb.

inc/Smartling/ApiWrapper.php
dimitrystd: >GDPR requirements

Yep, it appeared last week. I'm going to raise this question with Max and Evgeniy. Then, depending on discussions results, talk to AlexS.
But let's don't wait for decision and move on with user_id so far. We can add the real name later.

>setClientData (should we log any wp related data to the event like submission id/submission created/etc?)

Totally agree. Let's add IDs (page, source\target blogs, submission id)

inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -605,4 +608,10 @@ private function getSources(CloneRequest $request, int $targetBlogId): array
 
         return $sources;
     }
+
+    private function logSubmissionCreated(SubmissionEntity $submission, string $description, bool $isAuthorize): void
+    {
+        $this->getLogger()->info(sprintf("Adding submissionId=%d, fileName=%s, sourceBlogId=%d, sourceId=%d for upload to batchUid=%s due to user request, reason=\"%s\"", $submission->getId(), $submission->getStateFieldFileUri(), $submission->getSourceBlogId(), $submission->getSourceId(), $submission->getBatchUid(), $description));
```
dimitrystd, line 614: As for me, this log looks redundant. I think we should log once in `bulkUpload()` (before the loop).

inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -142,6 +143,7 @@ public function bulkUpload(JobEntityWithBatchUid $jobInfo, array $contentIds, st
                     $submission->setSourceTitle($title);
                 }
                 $this->submissionManager->storeEntity($submission);
+                $this->logSubmissionCreated($submission, 'Bulk upload request', false);
```
dimitrystd, line 146: I just realized that it will be 1 log entry per post and locale. It will be too much. We do similar for PP, but I'm not sure that we need it from the beginning here.
I thought about a single audit record for each user request. Very similar to what we send in ajax payload from UI (job, current post, auto-authorize; related items; etc).

dimitrystd (changes requested):

inc/Smartling/Services/ContentRelationsDiscoveryService.php
vsolovei-smartling: Ok, will change accordingly

inc/Smartling/ApiWrapper.php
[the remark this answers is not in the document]
vsolovei-smartling: Done

inc/Smartling/ApiWrapper.php
vsolovei-smartling: Done

inc/Smartling/Services/ContentRelationsDiscoveryService.php
vsolovei-smartling: Done

inc/Smartling/Services/ContentRelationsDiscoveryService.php

vsolovei-smartling:
I've also renamed CloneRequest and TranslationRequest to UserCloneRequest and UserTranslationRequest to clarify their intended use

inc/Smartling/Services/ContentRelationsDiscoveryService.php
vsolovei-smartling: I've extracted the audit log creation, but think no harm in having detailed submission creation logs

inc/Smartling/WP/Table/SubmissionTableWidget.php
```diff
@@ -186,9 +190,27 @@ private function processBulkAction(): void
...
+                            if ($profile === null) {
+                                $profile = $this->entityHelper->getSettingsManager()->getSingleSettingsProfile($submission->getSourceBlogId());
+                            }
+                            $logSubmissions[] = [
+                                'sourceBlogId' => $submission->getSourceBlogId(),
+                                'sourceId' => $submission->getSourceId(),
+                                'submissionId' => $submission->getId(),
+                                'targetBlogId' => $submission->getTargetBlogId(),
+                                'targetId' => $submission->getTargetId(),
+                            ];
```
PavelLoparev, line 205: Code duplications. Seems like could be extracted to some method

inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -199,13 +201,19 @@ public function clone(CloneRequest $request): void
...
         $profile = $this->settingsManager->getSingleSettingsProfile($curBlogId);
         $job = $request->getJobInformation();
         $batchUid = $this->getBatchUid($profile, $job);
         $jobInfo = new JobEntityWithBatchUid($batchUid, $job->getName(), $job->getId(), $profile->getProjectId());
+        $this->apiWrapper->createAuditLogRecord($profile, CreateRecordParameters::ACTION_TYPE_UPLOAD, $request->getDescription(), [
+            'relatedContentIds' => $request->getRelationsOrdered(),
+            'sourceBlogId' => $curBlogId,
+            'sourceId' => $request->getContentId(),
+            'targetBlogIds' => $request->getTargetBlogIds(),
+        ], $jobInfo, $job->isAuthorize());
```
PavelLoparev, line 216: @dimitrystd https://bt.smartling.net/browse/ALS-43

PavelLoparev (approved):

inc/Smartling/Services/ContentRelationsDiscoveryService.php
PavelLoparev: @vsolovei-smartling could you please validate in field types described in the ticket above?

inc/Smartling/WP/Table/SubmissionTableWidget.php
vsolovei-smartling: These are two different data sets with two different sources, added a comment in [bt](https://bt.smartling.net/browse/ALS-43)

dimitrystd (approved):

PR #375, opened 2022-04-26
add upload queue purge (WP-703)

vsolovei-smartling:


inc/Smartling/Queue/QueueInterface.php
```diff
@@ -2,16 +2,15 @@
...
- * @package Smartling\Queue
- */
 interface QueueInterface
 {
+    public const VIRTUAL_UPLOAD_QUEUE = 'upload-queue';
     public const QUEUE_NAME_DOWNLOAD_QUEUE = 'download-queue';
     public const QUEUE_NAME_LAST_MODIFIED_CHECK_AND_FAIL_QUEUE = 'last-modified-check-and-fail-queue';
     public const QUEUE_NAME_LAST_MODIFIED_CHECK_QUEUE = 'last-modified-check-queue';
 
+    public function isVirtual(string $queue): bool;
```
dimitrystd, line 12: PR approved but I don't like it. Maybe it's time to reconsider it. The absence of real queue for upload already has a drawback for us. We have column for jobBatchUid in table. But in general, this info should live in the context of upload flow only. So if I submit the page1 to job1 and then once again page1 to job2, then we should see 2 uploads into different jobs because it will be 2 upload messages.

P.S. This is how hosted connectors work and we don't see any drawbacks with this approach.

dimitrystd (approved):

inc/Smartling/Queue/Queue.php
```diff
@@ -98,6 +98,11 @@ private function testQueue($queue)
         return is_string($queue);
     }
 
+    public function isVirtual(string $queue): bool
+    {
+        return $queue === QueueInterface::VIRTUAL_UPLOAD_QUEUE;
+    }
```
PavelLoparev, line 104: Only upload queue can be virtual?

inc/Smartling/WP/Controller/ConfigurationProfilesController.php
```diff
@@ -111,59 +118,72 @@ public function processCnqAction()
...
+            case self::ACTION_QUEUE_PURGE:
+                try {
+                    if ($argument === QueueInterface::VIRTUAL_UPLOAD_QUEUE) {
+                        $submissions = $this->getManager()->find([
+                            SubmissionEntity::FIELD_STATUS => SubmissionEntity::SUBMISSION_STATUS_NEW,
+                        ]);
+                        if (count($submissions) > 0) {
+                            $ids = [];
+                            foreach ($submissions as $submission) {
+                                $submission->setStatus(SubmissionEntity::SUBMISSION_STATUS_CANCELLED);
```
PavelLoparev, line 134: Should we also provide an error message like `Cancelled while purging the queue` or something?

readme.txt
```diff
@@ -62,6 +62,10 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
+= 2.12.6 =
+* Added purge upload queue action (sets all NEW submissions to CANCELLED)
+* Fixed smartlingLockId attribute being sent for translation
```
PavelLoparev, line 67: I thought this was fixed in previous PR

inc/Smartling/Queue/QueueInterface.php
vsolovei-smartling: Created https://bt.smartling.net/browse/WP-706 to add proper queue

readme.txt
vsolovei-smartling: This was indeed in https://github.com/Smartling/wordpress-localization-plugin/pull/374, but I didn't release/add readme about it

PavelLoparev (approved):

PR #378, opened 2022-05-03
add meta values normalization for taxonomies (WP-704)

vsolovei-smartling:
The issue is that WordPress returns an array of values for get_term_meta, and we need to normalize this array just as is done for posts

### DIFF
--- a/tests/Smartling/DbAl/WordpressContentEntities/TaxonomyEntityStdTest.php
+++ b/tests/Smartling/DbAl/WordpressContentEntities/TaxonomyEntityStdTest.php
@@ -8,9 +8,9 @@ use Smartling\Helpers\WordpressFunctionProxyHelper;
 
 class TaxonomyEntityStdTest extends TestCase {
     /**
-     * Metadata values for taxonomies should not be stored as array for scalar values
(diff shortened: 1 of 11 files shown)


dimitrystd (approved):

tests/Smartling/DbAl/WordpressContentEntities/TaxonomyEntityStdTest.php
```diff
@@ -0,0 +1,25 @@
...
+
+use PHPUnit\Framework\TestCase;
+use Smartling\DbAl\WordpressContentEntities\TaxonomyEntityStd;
+use Smartling\Helpers\WordpressFunctionProxyHelper;
+
+class TaxonomyEntityStdTest extends TestCase {
+    /**
+     * Metadata values for taxonomies should not be stored as array for scalar values
+     */
+    public function testWP704()
```
PavelLoparev, line 13: Let's give meaningful names to tests. If it fails you will never recall what was the issue in WP-704 ticket. Let's put ticket reference in the test comment.

PavelLoparev (approved):

PR #379, opened 2022-05-05
add submission locked check for cloning (WP-709)

vsolovei-smartling:


dimitrystd (approved):

inc/Smartling/Submissions/SubmissionEntity.php
```diff
@@ -271,6 +271,11 @@ public function setWordCount(?int $word_count): void
         $this->stateFields[static::FIELD_WORD_COUNT] = $word_count;
     }
 
+    public function isLocked(): bool
```
dimitrystd, line 274: Strange that `getIsLocked`, `setIsLocked` still works with `int`.

vsolovei-smartling: All existing code depends on it, refactoring in this pr would cause much bloat

PR #380, opened 2022-05-12
add release version check (WP-712)

vsolovei-smartling:
It is possible to release a non-stable version, and the process of reverting such a release is non-trivial. I remember the idea of releasing by GH tag, but it needs more effort, and would be prone to the same mistake too

### DIFF
--- a/Buildplan/Dockerfile-Jenkins
+++ b/Buildplan/Dockerfile-Jenkins
@@ -67,9 +67,8 @@ ENV COMPOSER_H_DIR="${HOME}/.composer"
 ENV PLUGIN_DIR="${WP_PLUGINS_DIR}/smartling-connector"
 ENV LOCAL_GIT_DIR="/plugin-dir"
 
-COPY Buildplan/release.sh /release.sh
-COPY Buildplan/test.sh /test.sh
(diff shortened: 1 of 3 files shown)


Buildplan/Dockerfile-Jenkins
```diff
@@ -67,9 +67,8 @@ ENV COMPOSER_H_DIR="${HOME}/.composer"
 ENV PLUGIN_DIR="${WP_PLUGINS_DIR}/smartling-connector"
 ENV LOCAL_GIT_DIR="/plugin-dir"
 
-COPY Buildplan/release.sh /release.sh
-COPY Buildplan/test.sh /test.sh
-RUN chmod +x /release.sh /test.sh
+COPY Buildplan/release.sh Buildplan/releaseVersionCheck.sh Buildplan/test.sh /
+RUN chmod +x /release.sh /releaseVersionCheck.sh /test.sh
```
dimitrystd, line 71: Or use wildcard since it's our folder with limited number of files
```suggestion
RUN chmod +x /*.sh
```

dimitrystd (approved):

PavelLoparev (approved):

PR #385, opened 2022-05-19
catch Api error on audit log creation (WP-714)

vsolovei-smartling:
An exception in api caused uploads to fail, this should not happen


PavelLoparev (approved):

inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -213,12 +213,34 @@ public function createSubmissions(UserTranslationRequest $request): void
...
+                    }
+                }
+            }
+            foreach ($relatedIds as $contentType => $contentIds) {
+                $relatedIds[$contentType] = array_unique($contentIds);
+            }
+        } catch (\Exception $e) {
+            $this->getLogger()->error('Failed to build related ids array for audit log');
+        }
+        try {
```
dimitrystd, line 234: It's a good idea. Audit log must not interrupt any flow.
I would recommend to create a new class\service for audit log and catch errors\enrich by some date there. Example in [Paypal service](https://github.com/Smartling/paypal-github-service/blob/master/server/src/main/java/com/smartling/githubservice/service/sdk/AuditLogService.java).

PR #392, opened 2022-05-31
disable taxonomy handling with AIOSEO, fix field type issues (WP-726)

vsolovei-smartling:


inc/Smartling/ContentTypes/ExternalContentAioseo.php
```diff
@@ -75,6 +75,10 @@ public function __construct(SiteHelper $siteHelper, SmartlingToCMSDatabaseAccess
 
     public function getContentFields(SubmissionEntity $submission, bool $raw): array
     {
+        // Only process posts, because taxonomies are only supported in AIOSEO Pro
+        if (!in_array($submission->getContentType(), get_post_types(), true)) {
+            return [];
+        }
```
PavelLoparev, line 81: Can we check if current AIOSEO is `pro` or not? Seems like this change disables support even for `pro` version.

vsolovei-smartling: This check ensures that a submission is for post based content. I couldn't get the pro version for testing, so I'm not sure how that stores fields information, the basic version only has aioseo_posts table, values from that are not applicable to taxonomies. If we encounter a taxonomy submission, the function doesn't add any fields.

vsolovei-smartling: If we would need add support for taxonomy based content, this return statement would get replaced by the taxonomy table processing

PavelLoparev (approved):

dimitrystd (approved):

PR #393, opened 2022-06-08
add ACF related items when sending (WP-730)

vsolovei-smartling:


inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -410,41 +415,38 @@ private function addFlattenedBlock(array $array, GutenbergBlock $block, string $
...
+     */
     private function getPostContentReferences(string $content): array
     {
         $result = [];
-        foreach ($this->gutenbergBlockHelper->parseBlocks($content) as $block) {
-            $result = $this->addPostContentReferences($result, $block);
+        try {
+            foreach ($this->gutenbergBlockHelper->parseBlocks($content) as $block) {
+                $result = $this->addPostContentReferences($result, $block);
+            }
```
PavelLoparev, line 427: Maybe wrap into try/catch exact call but not the entire loop? So that if only one parse failed others will still be parsed

PavelLoparev (approved):
Approved but did not see tests for `related items from within ACF Gutenberg blocks`

PavelLoparev (approved):

dimitrystd (approved):

PR #402, opened 2022-07-08
change logging (WP-739)

vsolovei-smartling:


inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -88,6 +88,9 @@ private function setFileUriIfNullId(SubmissionEntity $submission): SubmissionEnt
     private function createTargetContent(SubmissionEntity $submission): SubmissionEntity
     {
         $submission = $this->getFunctionProxyHelper()->apply_filters(ExportedAPI::FILTER_SMARTLING_PREPARE_TARGET_CONTENT, $submission);
+        if (!$submission instanceof SubmissionEntity) {
+            $this->getLogger()->critical('Submission not instance of ' . SubmissionEntity::class . ' after filter ' . ExportedAPI::FILTER_SMARTLING_PREPARE_TARGET_CONTENT);
+        }
 
         if (SubmissionEntity::SUBMISSION_STATUS_FAILED === $submission->getStatus()) {
```
PavelLoparev, line 95: Should you stop execution if it's not of expected type? Because for now if it's not it will still call `$submission->getStatus()` (getStatus might not exist on $submission)? Same comment for changes in other places

vsolovei-smartling: It will fail with an exception, and might get into the logs, this logging purpose is to be able to 100% pinpoint the issue is with clients' hooks

vsolovei-smartling: I've added that statement to the log body, this way it should be clear for anyone reading the logs

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -88,6 +88,9 @@ private function setFileUriIfNullId(SubmissionEntity $submission): SubmissionEnt
     private function createTargetContent(SubmissionEntity $submission): SubmissionEntity
     {
         $submission = $this->getFunctionProxyHelper()->apply_filters(ExportedAPI::FILTER_SMARTLING_PREPARE_TARGET_CONTENT, $submission);
+        if (!$submission instanceof SubmissionEntity) {
```
dimitrystd, line 91: Maybe make it more explicit by adding brackets. I will not be surprised if PHP will invert value of `$submission` and then will check its type
```suggestion
        if (!($submission instanceof SubmissionEntity)) {
```

inc/Smartling/DbAl/WordpressContentEntities/PostEntityStd.php
```diff
@@ -239,6 +239,7 @@ public function set(EntityAbstract $entity = null)
         remove_action('content_save_pre', 'acf_parse_save_blocks', 5);
         if (isset($array['post_content'])) {
             // Content needs to be slashed for wp_insert_post() call
+            $this->getLogger()->info(sprintf('Saving postId=%d, content: "%s"', $array['ID'] ?? 0, base64_encode($array['post_content'])));
```
dimitrystd, line 242: Current logs already noisy. The new one will add entire post in log (what we already had in other places). Can we do it more specific? Write only if some condition, or write what exactly wrong.

inc/Smartling/Base/SmartlingCoreUploadTrait.php
vsolovei-smartling: ! has lower precedence than instanceof, but explicit is better

inc/Smartling/DbAl/WordpressContentEntities/PostEntityStd.php
vsolovei-smartling: Removing this for now

PR #408, opened 2022-07-27
fix metadata escaping (WP-747)

vsolovei-smartling:
Previously, a fix was applied to lessen the amount of escaping-related issues while saving post_content field. It worked, and now it appears we need to rollout the same for meta fields and other relevant post fields

### DIFF
--- a/inc/Smartling/DbAl/WordpressContentEntities/PostEntityStd.php
+++ b/inc/Smartling/DbAl/WordpressContentEntities/PostEntityStd.php
@@ -3,7 +3,6 @@
 namespace Smartling\DbAl\WordpressContentEntities;
 
 use Smartling\Exception\SmartlingDataUpdateException;
-use Smartling\Helpers\ArrayHelper;
 use Smartling\Helpers\RawDbQueryHelper;
 use Smartling\Helpers\WordpressFunctionProxyHelper;
 use Smartling\Helpers\WordpressUserHelper;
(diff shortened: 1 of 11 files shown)


inc/Smartling/DbAl/WordpressContentEntities/PostEntityStd.php
```diff
@@ -237,9 +236,15 @@ public function set(EntityAbstract $entity = null)
...
         remove_action('content_save_pre', 'acf_parse_save_blocks', 5);
-        if (isset($array['post_content'])) {
-            // Content needs to be slashed for wp_insert_post() call
-            $array['post_content'] = addslashes($array['post_content']);
+
+        /**
+         * Content expected to be slashed for
+         * @see wp_insert_post() $data declaration
+         */
+        foreach (['post_author', 'post_content', 'post_content_filtered', 'post_title', 'post_excerpt', 'post_password', 'post_name', 'to_ping', 'pinged'] as $field) {
```
PavelLoparev, line 244: Can this list be potentially extended with some properties from external plugins? If so then can we introduce a hook/action or something that will allow us/developers to "register" new properties instead of hardcoding them here?

vsolovei-smartling: No, this is inner WordPress workings, no potential extension except for major WordPress updates, I believe

PR #409, opened 2022-07-28
fix post content not sent for translation if Elementor is active but not used for a specific post (WP-750)

vsolovei-smartling:


inc/Smartling/ContentTypes/ExternalContentAbstract.php
```diff
@@ -3,12 +3,16 @@
...
 
 abstract class ExternalContentAbstract implements ContentTypePluggableInterface {
-    public function canHandle(PluginHelper $pluginHelper): bool
+    public function canHandle(PluginHelper $pluginHelper, int $contentId, ?WordpressFunctionProxyHelper $wpProxy = null): bool
     {
-        $activePlugins = wp_get_active_network_plugins();
-        $plugins = get_plugins();
+        if ($wpProxy === null) {
+            $wpProxy = new WordpressFunctionProxyHelper();
+        }
```
PavelLoparev, line 13: Looks like a hack. Maybe let's move `?WordpressFunctionProxyHelper $wpProxy = null` to the `ContentTypePluggableInterface` interface and update all the places where `canHandle` is called in order to eventually pass the `WordpressFunctionProxyHelper` instance?

inc/Smartling/Helpers/WordpressFunctionProxyHelper.php
```diff
@@ -47,19 +54,14 @@ public function delete_post_meta()
...
-     * @param int $blogId
-     * @param int $postId
-     * @return mixed
-     */
-    public function get_blog_permalink(int $blogId, int $postId)
+    public function get_plugins(): array
     {
-        return get_blog_permalink(...func_get_args());
+        return get_plugins(...func_get_args());
     }
```
PavelLoparev, line 65: Are we sure `delete_term_meta` and `get_blog_permalink` are never used in other places?

tests/Smartling/ContentTypes/ExternalContentElementorTest.php
```diff
@@ -5,11 +5,24 @@
...
+    public function testCanHandle()
+    {
+        $pluginHelper = $this->createMock(PluginHelper::class);
+        $pluginHelper->method('versionInRange')->willReturn(true);
+        $proxy = $this->createMock(WordpressFunctionProxyHelper::class);
+        $proxy->method('getPostMeta')->willReturn('', []);
+        $proxy->method('get_plugins')->willReturn(['elementor/elementor.php' => []]);
+        $proxy->method('wp_get_active_network_plugins')->willReturn(['elementor/elementor.php']);
+        $this->assertFalse($this->getExternalContentElementor($proxy)->canHandle($pluginHelper, 1, $proxy));
+        $this->assertTrue($this->getExternalContentElementor($proxy)->canHandle($pluginHelper, 1, $proxy));
```
PavelLoparev, line 23: Didn't get why the second consecutive call worked.

inc/Smartling/ContentTypes/ExternalContentAbstract.php
PavelLoparev: It seems like instance of `WordpressFunctionProxyHelper` **must** be passed here anyway?

vsolovei-smartling: It is used for testing, not sure if we need to  move it to the interface.

inc/Smartling/Helpers/WordpressFunctionProxyHelper.php
vsolovei-smartling: Yes. The only deleted function was delete_term_meta

tests/Smartling/ContentTypes/ExternalContentElementorTest.php
vsolovei-smartling: On line 19 here we setup our proxy to return empty meta (key does not exist or empty) for our key on the first call, so the handler knows it can't handle. The second call returns an empty array for the key, and the handler knows it can handle

inc/Smartling/ContentTypes/ExternalContentAbstract.php
vsolovei-smartling: Done

PavelLoparev (approved):

inc/Smartling/ContentTypes/ExternalContentElementor.php
```diff
@@ -136,6 +137,12 @@ public function alterContentFieldsForUpload(array $source): array
         return $source;
     }
 
+    public function canHandle(PluginHelper $pluginHelper, WordpressFunctionProxyHelper $wpProxy, int $contentId): bool
```
dimitrystd, line 140: @vsolovei-smartling I think it's the first step of screwing up the nice idea of a single interface for all integrations.
Why you don't analyze post in `getContentFields` or `extractContent`?  These methods can always return null-object and everything else must work as expected.

vsolovei-smartling: Not sure what you mean by that. We call getContentFields and expect it to return array. We could move the check to any of the methods, but what would change? Now we have a canHandle that determines whether to call getContentFields. Would you prefer to change getContentFields to return the same array if content isn't supported? Currently if someone were to call getContentFields without checking for canHandle, an exception would be thrown, but I'm for this behaviour.

PR #410, opened 2022-08-02
add beaver builder support (WP-744)

vsolovei-smartling:


inc/Smartling/ContentTypes/ExternalContentAbstract.php
```diff
@@ -2,11 +2,16 @@
...
 use Smartling\Helpers\PluginHelper;
 use Smartling\Helpers\WordpressFunctionProxyHelper;
+use Smartling\Submissions\SubmissionEntity;
+use Smartling\Submissions\SubmissionManager;
 
 abstract class ExternalContentAbstract implements ContentTypePluggableInterface {
-    public function canHandle(PluginHelper $pluginHelper, WordpressFunctionProxyHelper $wpProxy, int $contentId): bool
+    use LoggerSafeTrait;
+
+    public function canHandle(ContentTypeHelper $contentTypeHelper, PluginHelper $pluginHelper, WordpressFunctionProxyHelper $wpProxy, int $contentId, string $contentType): bool
```
vsolovei-smartling, line 14: @dimitrystd, let's return to the discussion https://github.com/Smartling/wordpress-localization-plugin/pull/409#discussion_r933628075

inc/Smartling/ContentTypes/ContentTypePluggableInterface.php
```diff
@@ -8,7 +8,7 @@
 
 interface ContentTypePluggableInterface
 {
-    public function canHandle(PluginHelper $pluginHelper, WordpressFunctionProxyHelper $wpProxy, int $contentId): bool;
+    public function canHandle(ContentTypeHelper $contentTypeHelper, PluginHelper $pluginHelper, WordpressFunctionProxyHelper $wpProxy, int $contentId, string $contentType): bool;
```
dimitrystd, line 11: I don't think standard dependencies should be part of the contract. You can pass all of these helpers in class constructor. I can't imagine the case when you need different `$contentTypeHelper` implementations for different `$contentId`.

inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
```diff
@@ -0,0 +1,338 @@
...
+        return $translation;
+    }
+
+    private function buildData(array $original, array $translation, SubmissionEntity $submission): array
+    {
+        $result = $original;
+
+        foreach ($original as $key => $value) {
+            if (array_key_exists($key, $translation) && property_exists($value, 'settings')) {
+                if (array_key_exists('settings', $translation[$key])) {
```
PavelLoparev, line 144: Let's merge to one condition?

inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
```diff
@@ -0,0 +1,338 @@
...
+                    $arrayOriginalSettings = (array)$value->settings;
+                    foreach ($arrayOriginalSettings as $settingKey => $setting) {
+                        if (array_key_exists($settingKey, $translation[$key]['settings'])) {
+                            if (is_scalar($translation[$key]['settings'][$settingKey])) {
+                                settype($translation[$key]['settings'][$settingKey], gettype($setting));
```
PavelLoparev, line 149: same here

inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
```diff
@@ -0,0 +1,338 @@
...
+    private function flattenObject(\stdClass $object, string $base, string $divider): array
+    {
+        $result = [];
+        foreach ((array)$object as $key => $value) {
+            $path = '' === $base ? $key : implode($divider, [$base, $key]);
+            $result[] = $this->processArrayElement($path, $value, $divider);
+        }
+
+        return array_merge(...$result);
+    }
```
PavelLoparev: Identical methods. why not just cast object to array before passing it to `flatten`?

inc/Smartling/ContentTypes/ContentTypePluggableInterface.php
vsolovei-smartling: Done

inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
vsolovei-smartling: Weird, this was done before you commented

inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
vsolovei-smartling: Can't be done, there's an alternative branch at line 166 that also needs to be checked if the condition at line 142 is true

inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
vsolovei-smartling: Again, can't be done, two different branches that depend on parent check

inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
```diff
@@ -0,0 +1,328 @@
...
+    public function getContentFields(SubmissionEntity $submission, bool $raw): array
+    {
+        $submission->assertHasSource();
+        return $this->extractContent($this->getDataFromPostMeta($submission->getSourceId()))->getStrings();
+    }
+
+    private function getDataFromPostMeta(int $id): array
+    {
+        return $this->wpProxy->getPostMeta($id, self::META_FIELD_NAME, true);
+    }
```
PavelLoparev, line 103: Can this (and other methods that call this one) be moved to parent `ExternalContentAbstract` class?

Seems like the only change would be `self::META_FIELD_NAME` -> `static::META_FIELD_NAME` (use late static binding so that in child class you just define `META_FIELD_NAME`)

PavelLoparev: I mean seems like this method will be copypsated for every new integration

inc/Smartling/ContentTypes/ExternalContentAbstract.php
```diff
@@ -32,4 +46,29 @@ public function getRelatedContent(string $contentType, int $id): array
     {
         return [];
     }
+
+    protected function getTargetAttachmentId(SubmissionManager $submissionManager, SubmissionEntity $submission, int $attachmentId): ?int
```
dimitrystd, line 50: Thanks for updating `ContentTypePluggableInterface`. The `$submissionManager` is the same use case. If class needs it, then set in constructor. It can be in methods parameters only if its implementation will be different when `getTargetAttachmentId` is called.

inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
vsolovei-smartling: I moved getDataFromPostMeta to parent, but it can't be done for the calling methods, even between those two integrations, one unserializes content, the other json_decodes

inc/Smartling/ContentTypes/ExternalContentAbstract.php
vsolovei-smartling: Done, although this does mean that ExternalContentAioseo now has an unneeded dependency for SubmissionManager. Probably, most of the integrations will need SubmissionManager, so, I'm ok with that

PavelLoparev (approved):

dimitrystd (approved):
Looks much cleaner

PR #415, opened 2022-08-09
allow null values for StringHelper string shrink functions (WP-754)

vsolovei-smartling:
metadata might be null, causing critical errors

### DIFF
--- a/inc/Smartling/Helpers/StringHelper.php
+++ b/inc/Smartling/Helpers/StringHelper.php
@@ -27,8 +27,11 @@ class StringHelper
         );
     }
 
-    public static function safeHtmlStringShrink(string $string, int $maxLength = 50, string $encoding = 'utf8', bool $applyHtmlEncoding = true, string $wrapperTag = 'span'): string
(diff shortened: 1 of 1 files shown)


inc/Smartling/Helpers/StringHelper.php
```diff
@@ -27,8 +27,11 @@ public static function buildPattern(string $pattern, string $borders = '/', stri
         );
     }
 
-    public static function safeHtmlStringShrink(string $string, int $maxLength = 50, string $encoding = 'utf8', bool $applyHtmlEncoding = true, string $wrapperTag = 'span'): string
+    public static function safeHtmlStringShrink(?string $string, int $maxLength = 50, string $encoding = 'utf8', bool $applyHtmlEncoding = true, string $wrapperTag = 'span'): string
     {
+        if ($string === null) {
+            return '';
```
dimitrystd, line 33: Not sure where do you plan to use it. But are you sure that converting `null` to empty string will be ok for business logic?

dimitrystd (approved):

inc/Smartling/Helpers/StringHelper.php
vsolovei-smartling: It is only used when displaying information on frontend, and a null value is ok to display as empty string. This would have to be changed when migrating to php8, but due to a user experiencing issue with it, it has to be done earlier

PR #418, opened 2022-09-01
add expert settings to add slashes to post content and metadata (WP-731)

vsolovei-smartling:


inc/Smartling/DbAl/WordpressContentEntities/PostEntityStd.php
```diff
@@ -241,11 +242,16 @@ public function set(EntityAbstract $entity = null)
...
+        $addSlashes = GlobalSettingsManager::isAddSlashesBeforeSavingPostContent();
+        if ($addSlashes) {
+            foreach (['post_author', 'post_content', 'post_content_filtered', 'post_title', 'post_excerpt', 'post_password', 'post_name', 'to_ping', 'pinged'] as $field) {
+                if (isset($array[$field])) {
+                    $array[$field] = addslashes($array[$field]);
+                }
             }
         }
+        /** @noinspection JsonEncodingApiUsageInspection failure to json_encode suitable for logging */
+        $this->getLogger()->debug(sprintf('Calling wp_insert_post with postArray="%s", addSlashes=%s', json_encode($array), $addSlashes));
```
vsolovei-smartling, line 254: This would prove useful when debugging issues, I still believe that with the later plugin versions only escaping issues arise from third party plugins or custom theme code.

inc/Smartling/DbAl/WordpressContentEntities/PostEntityStd.php
```diff
@@ -369,18 +375,22 @@ public function setMetaTag($tagName, $tagValue, $unique = true): void
...
+                $tagValue = addslashes($tagValue);
+            }
         }
         if (false === ($result = add_post_meta($this->ID, $tagName, $tagValue, $unique))) {
             $result = update_post_meta($this->ID, $tagName, $tagValue);
         }
 
         if (false === $result) {
-            if (false === $this->ensureMetaValue($tagName, $tagValue)) {
+            if (false === $this->ensureMetaValue($expectedTagName, $expectedTagValue)) {
```
vsolovei-smartling, line 393: Fix not strictly related to the issue: adding escaping to metadata throws the check off, generating log errors when metadata is present

inc/Smartling/WP/View/ConfigurationProfiles.php
```diff
@@ -56,7 +56,7 @@
             <a href="<?= get_site_url(); ?>/wp-admin/admin-post.php?action=smartling_zerolength_log_file"><?= __('DELETE current log file.'); ?></a>
         </li>
 
-        <div class="update-nag">
```
vsolovei-smartling, line 59: Not strictly related to the issue, a user complained that they are unable to see the expert settings link entirely. This is most likely due to a "nag-disabler" plugin in WP

PavelLoparev (commented):
Do I understand correctly that our plugin always added slashes but there are other plugins that do the same before saving content? Sound like a problem. Nobody will remember that when those plugins are disabled then somebody would need to enable this extra settings in connector.

vsolovei-smartling:
> Do I understand correctly that our plugin always added slashes but there are other plugins that do the same before saving content? Sound like a problem. Nobody will remember that when those plugins are disabled then somebody would need to enable this extra settings in connector.

Generally, yes, it's on by default and would be needed to be turn off in case someone sees unnecessary slashes in the post content or metadata.

PavelLoparev (approved):

PR #422, opened 2022-09-15
add notice about unsupported versions for external content handlers (WP-762)

vsolovei-smartling:


inc/config/services.yml
```diff
@@ -139,10 +141,6 @@ services:
 
   manager.content.external:
     class: Smartling\ContentTypes\ExternalContentManager
-    arguments:
-      - "@content.type.helper"
-      - "@helper.plugins"
-      - "@wp.proxy"
```
PavelLoparev, line 145: Not used?

inc/Smartling/Bootstrap.php
```diff
@@ -273,6 +278,39 @@ private function testCronSetup(): void
...
+            foreach ($handlers as $handler) {
+                if (!$wpProxy->is_plugin_active($pluginPath) || $pluginHelper->pluginVersionInRange($pluginPath, $handler->getMinVersion(), $handler->getMaxVersion())) {
+                    continue 2;
+                }
+            }
+            $unsupportedPluginPaths[] = $pluginPath;
+        }
+        if (count($unsupportedPluginPaths) > 0) {
+            DiagnosticsHelper::addDiagnosticsMessage('Unsupported plugin version detected: ' . implode(', ', $unsupportedPluginPaths));
+        }
```
PavelLoparev, line 311: Didn't get. Test pass if no exceptions were thrown?

This call `DiagnosticsHelper::addDiagnosticsMessage` doesn't seem to do anything with the test result? I mean do we need to assert these added messages somehow?

inc/config/services.yml
vsolovei-smartling: Not used

inc/Smartling/Bootstrap.php
vsolovei-smartling: Exceptions are for badly misconfigured installations. The container is hardcoded for these services, but in case something drastically changes, these checks are introduced. The result of addDiagnosticMessage is used at https://github.com/Smartling/wordpress-localization-plugin/blob/master/inc/Smartling/Helpers/UiMessageHelper.php#L19, to display the messages. I've further tweaked the output to indicate that the source of the messages is the Smartling plugin, they are ambiguous now.

vsolovei-smartling: image
This is what the updated messages would look like

PavelLoparev (approved):

dimitrystd (approved):

PR #433, opened 2022-11-11
use plugin registered content types for relations discovery (WP-775)

vsolovei-smartling:


tests/Smartling/WP/Table/SubmissionTableWidgetTest.php
```diff
@@ -68,7 +68,14 @@ public function testBulkSubmitDownload(): void
...
                 $queue,
-            ])->addMethods(['current_action'])->getMock();
+            ]);
+            // Is this running with WordPress classes?
+            if (method_exists('WP_List_Table', 'current_action')) {
+                $x->onlyMethods(['current_action']);
+            } else {
+                $x->addMethods(['current_action']);
+            }
+            $x = $x->getMock();
```
vsolovei-smartling, line 78: A warning was introduced in a previous release, and I didn't notice until now. When running only the unit test suite, addMethods is proper because no WP_List_Table class exists, but when running both suites, onlyMethods needs to be used to avoid a warning

dimitrystd (approved):

PavelLoparev (approved):

PR #423, opened 2022-09-21
migrate to php8 (wp-751)

vsolovei-smartling:
I initially planned to only remove warnings, but since php 7.4 security support will end on 28 November, we might as well migrate to 8.1 now

PR #435, opened 2022-11-21
allow beaver builder 2.6

vsolovei-smartling:


Jenkinsfile
```diff
@@ -18,39 +18,39 @@ pipeline {
...
             agent {
                 label 'master'
             }
 
             steps {
-                withCredentials([string(credentialsId: 'PROJECT_ID', variable: 'PROJECT_ID'), string(credentialsId: 'USER_IDENTIFIER', variable: 'USER_IDENTIFIER'), string(credentialsId: 'TOKEN_SECRET', variable: 'TOKEN_SECRET')]) {
-                    dir('.') {
-                        sh 'docker run --rm -w /plugin-dir -v $PWD:/plugin-dir -e MYSQL_HOST=localhost -e CRE_PROJECT_ID=$PROJECT_ID -e CRE_USER_IDENTIFIER=$USER_IDENTIFIER -e CRE_TOKEN_SECRET="$TOKEN_SECRET" wordpress-localization-plugin-php74:latest'
-                    }
+                dir('.') {
```
PavelLoparev, line 27: Is it needed in this case? Seems like can be omitted since it's current directory

composer.json
```diff
@@ -1,12 +1,18 @@
...
+  "repositories": [
+    {
+      "type": "github",
+      "url": "https://github.com/vsolovei-smartling/retry"
+    }
+  ],
   "require": {
     "galbar/jsonpath": "^2.0",
-    "jralph/retry": "^2.0",
+    "vsolovei-smartling/retry": "^2.0.1",
```
PavelLoparev, line 15: I thought we already use your lib. i remember PR in the past

PavelLoparev (approved):

Jenkinsfile
vsolovei-smartling: Removed, thanks

composer.json
vsolovei-smartling: I'll revert the change for now, this actually won't work without #423

PR #436, opened 2022-11-23
add exception handling in UploadJob (WP-776)

vsolovei-smartling:
There is logging in JobAbstract, but it would help to fail submissions, so the user wouldn't need to purge the queue manually. Also, a failure reason would persist in submission.

### DIFF
--- a/inc/Smartling/Jobs/UploadJob.php
+++ b/inc/Smartling/Jobs/UploadJob.php
@@ -50,7 +50,7 @@ class UploadJob extends JobAbstract
 
             try {
                 do_action(ExportedAPI::ACTION_SMARTLING_SEND_FILE_FOR_TRANSLATION, $entity);
-            } catch (\Error $e) {
+            } catch (\Exception $e) {
                 $this->getLogger()->notice(sprintf('Failing submissionId=%s: %s', $entity->getId(), $e->getMessage()));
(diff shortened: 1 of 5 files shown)


dimitrystd (approved):

inc/Smartling/Jobs/UploadJob.php
```diff
@@ -55,9 +48,14 @@ private function processUploadQueue(): void
                 )
             );
 
-            do_action(ExportedAPI::ACTION_SMARTLING_SEND_FILE_FOR_TRANSLATION, $entity);
+            try {
+                do_action(ExportedAPI::ACTION_SMARTLING_SEND_FILE_FOR_TRANSLATION, $entity);
+            } catch (\Exception $e) {
+                $this->getLogger()->notice(sprintf('Failing submissionId=%s: %s', $entity->getId(), $e->getMessage()));
+                $this->submissionManager->setErrorMessage($entity, $e->getMessage());
```
dimitrystd, line 55: If a post was requested for 3 langs then we will have in DB 3 records. But old\new code gets only one submission. So if something is wrong then this code will try to upload the same post 3 times (because you fail only a single submission).
BTW How do we update submission(s) if upload is successful? I think something like this should be done here.
Maybe `ACTION_SMARTLING_SEND_FILE_FOR_TRANSLATION` should accept a collection of "related" submission entries.

PR #444, opened 2023-01-20
refactor ContentRelationDiscoveryService (WP-786)

vsolovei-smartling:
there was no usage of field names, yet they were messing with the later discovered ids of post-based content

### DIFF
--- a/inc/Smartling/Services/ContentRelationsDiscoveryService.php
+++ b/inc/Smartling/Services/ContentRelationsDiscoveryService.php
@@ -545,7 +545,7 @@ class ContentRelationsDiscoveryService
                 if ($processor instanceof MetaFieldProcessorAbstract && 0 !== (int)$fValue) {
                     $shortProcessorName = ArrayHelper::last(explode('\\', get_class($processor)));
 
-                    $detectedReferences[$shortProcessorName][] = (int)$fValue; // TODO uncomment fix, remove following line
+                    $detectedReferences[$shortProcessorName][] = (int)$fValue;
                 } else {
                     $detectedReferences['attachment'] = array_merge($detectedReferences['attachment'],
                         $this->absoluteLinkedAttachmentCoreHelper->getImagesIdsFromString($fValue, $curBlogId));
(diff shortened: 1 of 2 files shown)


PavelLoparev (approved):

inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -544,24 +545,19 @@ public function getRelations(string $contentType, int $id, array $targetBlogIds)
                 if ($processor instanceof MetaFieldProcessorAbstract && 0 !== (int)$fValue) {
                     $shortProcessorName = ArrayHelper::last(explode('\\', get_class($processor)));
 
-                    $detectedReferences[$shortProcessorName][$fValue][] = $fName;
+                    $detectedReferences[$shortProcessorName][] = (int)$fValue; // TODO uncomment fix, remove following line
```
PavelLoparev, line 548: 1. Didn't get this TODO comment. There is nothing to uncomment.
2. I didn't get how it worked before then

dimitrystd (approved):
Also didn't get these changes. 🤞 that it will work. The `getRelations` is too long and complex

inc/Smartling/Services/ContentRelationsDiscoveryService.php
vsolovei-smartling: Comment removed, thanks
Consider the now-deleted [testNormalizeReferences](https://github.com/Smartling/wordpress-localization-plugin/pull/444/files?diff=split&w=0#diff-e6ee028c484376b4e345d0921c38841681d3ad8c83d9faffe7b94bf8bc78ade1L659). Previously, when a MetaFieldProcessorAbstract descendant added an entry, it looked like
```
$detectedReferences['someProcessor'][someIntegerId] = 'someStringFieldName';
```
then, while detecting the post content references, the later processor would try to do
```
$detectedReferences[] = someIntegerId
```
because that was what is expected after normalisation -- an $detectedReferences should be an int[].
The normalisation [used to work with the keys of the array, however](https://github.com/Smartling/wordpress-localization-plugin/blob/12f0612fc01483e022454ed124fc53ad420cc8b5/inc/Smartling/Services/ContentRelationDiscoveryService.php#L546), until I've changed the function to work around the issue of adding items to an array that already has an integer key.

PavelLoparev (approved):

PR #446, opened 2023-01-27
add Gravity Forms support (WP-782)

vsolovei-smartling:
This is a first external plugin that needed to add a custom post type for processing. Extending EntityAbstract was the way to go, but it's design is bad. I've tried to improve upon it by creating interfaces for Entity and EntityHandler, and also created EntityWithMetadata and EntityWithPostStatus, to reduce the required effort to add new entities.

### DIFF
--- a/inc/Smartling/Base/SmartlingCore.php
+++ b/inc/Smartling/Base/SmartlingCore.php
@@ -5,6 +5,7 @@ namespace Smartling\Base;
 use Exception;
 use Smartling\ContentTypes\ContentTypeNavigationMenu;
 use Smartling\ContentTypes\ExternalContentManager;
+use Smartling\DbAl\WordpressContentEntities\EntityAbstract;
 use Smartling\Exception\BlogNotFoundException;
 use Smartling\Exception\SmartlingDbException;
 use Smartling\Exception\SmartlingExceptionAbstract;
@@ -70,7 +71,7 @@ class SmartlingCore extends SmartlingCoreAbstract
             $submission->getId(),
         ]));
         $originalEntity = $this->getContentHelper()->readSourceContent($submission);
-        $relatedContentTypes = $originalEntity->getRelatedTypes();
+        $relatedContentTypes = $originalEntity instanceof EntityAbstract ? $originalEntity->getRelatedTypes() : [];
         $accumulator = [
             'category' => [],
             'post_tag' => [],
--- a/inc/Smartling/Base/SmartlingCore.php
+++ b/inc/Smartling/Base/SmartlingCore.php
@@ -71,7 +71,7 @@ class SmartlingCore extends SmartlingCoreAbstract
             $submission->getId(),
         ]));
         $originalEntity = $this->getContentHelper()->readSourceContent($submission);
-        $relatedContentTypes = $originalEntity instanceof EntityAbstract ? $originalEntity->getRelatedTypes() : [];
+        $relatedContentTypes = $originalEntity->getRelatedTypes();
         $accumulator = [
             'category' => [],
             'post_tag' => [],
--- a/inc/Smartling/Base/SmartlingCoreAttachments.php
+++ b/inc/Smartling/Base/SmartlingCoreAttachments.php
@@ -2,6 +2,7 @@
 
 namespace Smartling\Base;
 
+use JetBrains\PhpStorm\ArrayShape;
 use Smartling\Exception\SmartlingWpDataIntegrityException;
 use Smartling\Helpers\AttachmentHelper;
 use Smartling\Helpers\StringHelper;
@@ -83,7 +84,7 @@ trait SmartlingCoreAttachments
      *
      * @param SubmissionEntity $submission
      */
-    public function regenerateTargetThumbnailsBySubmission(SubmissionEntity $submission)
+    public function regenerateTargetThumbnailsBySubmission(SubmissionEntity $submission): void
     {
         $this->getLogger()->debug(
             vsprintf('Starting thumbnails regeneration for blog = \'%s\' attachment id = \'%s\'.', [
@@ -108,42 +109,20 @@ trait SmartlingCoreAttachments
 
         $metadata = wp_generate_attachment_metadata($submission->getTargetId(), $originalImage);
 
-        if (is_wp_error($metadata)) {
-
-            $this->getLogger()
-                ->error(vsprintf('Error occurred while regenerating thumbnails for blog=\'%s\' attachment id=\'%s\'. Message:\'%s\'.', [
-                    $submission->getTargetBlogId(),
-                    $submission->getTargetId(),
-                    $metadata->get_error_message(),
-                ]));
-            $this->getLogger()
-                ->error(var_export($metadata, true));
-
-        }
-
         if (empty($metadata)) {
             $this->getLogger()
-                ->error(vsprintf('Couldn\'t regenerate thumbnails for blog=\'%s\' attachment id=\'%s\'.', [
+                ->error(sprintf("Couldn't regenerate thumbnails for blog='%s' attachment id='%s'.",
                     $submission->getTargetBlogId(),
                     $submission->getTargetId(),
-                ]));
+                ));
         } else {
-
-
             wp_update_attachment_metadata($submission->getTargetId(), $metadata);
-
         }
-
 
         $this->getContentHelper()->ensureRestoredBlogId();
     }
 
-    /**
-     * @param int $siteId
-     *
-     * @return array
-     */
-    private function getUploadDirForSite($siteId)
+    private function getUploadDirForSite(int $siteId): array
     {
         $this->getContentHelper()->ensureBlog($siteId);
         $data = wp_upload_dir();
@@ -152,7 +131,10 @@ trait SmartlingCoreAttachments
         return $data;
     }
 
-    private function getUploadPathForSite($siteId)
+    /** @noinspection PhpUnusedPrivateMethodInspection
+     * @see SmartlingCoreExportApi::getFullyRelateAttachmentPathByBlogId
+     */
+    private function getUploadPathForSite(int $siteId): string
     {
         $this->getContentHelper()->ensureBlog($siteId);
         $prefix = $this->getUploadDirForSite($siteId);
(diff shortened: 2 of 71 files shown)


inc/Smartling/Base/SmartlingCoreAttachments.php
```diff
@@ -108,42 +109,20 @@ public function regenerateTargetThumbnailsBySubmission(SubmissionEntity $submiss
 
         $metadata = wp_generate_attachment_metadata($submission->getTargetId(), $originalImage);
 
-        if (is_wp_error($metadata)) {
```
vsolovei-smartling, line 111: wp_generate_attachment_metadata always returns array

inc/Smartling/Base/SmartlingCoreAttachments.php
```diff
@@ -162,17 +144,27 @@ private function getUploadPathForSite($siteId)
...
-     *
-     * @return array
      * @throws SmartlingWpDataIntegrityException
      */
-    private function getAttachmentFileInfoBySubmission(SubmissionEntity $submission)
+    private function getAttachmentFileInfoBySubmission(SubmissionEntity $submission): array
     {
         $info = $this->getContentHelper()->readSourceContent($submission);
+        if (property_exists($info, 'guid')) {
+            $guid = $info->guid;
```
vsolovei-smartling, line 163: This was a magical method read from EntityAbstract, guid is always present for built-in WordPress content, but no substitute in the form of an uri for external content. This is used to detect attachments by uri, and should be instead handled by the relevant external handler

inc/Smartling/Base/SmartlingCoreTrait.php
```diff
@@ -58,7 +60,9 @@ protected function prepareTargetEntity(SubmissionEntity $submission, bool $force
 
         if (false === $update) {
             $targetContent = clone $originalContent;
-            $targetContent->cleanFields();
+            if ($targetContent instanceof EntityAbstract) {
+                $targetContent->cleanFields();
```
vsolovei-smartling, line 64: This removes all fields that should not be cloned. External handlers are/should be written in a way that only include fields to work with translation, so this method would not be needed

inc/Smartling/Base/SmartlingCoreTrait.php
```diff
@@ -73,6 +77,7 @@ protected function prepareTargetEntity(SubmissionEntity $submission, bool $force
         unset ($hardFilteredOriginalData['entity']['ID'], $hardFilteredOriginalData['entity']['term_id'], $hardFilteredOriginalData['entity']['id']);
 
         if (array_key_exists('entity', $hardFilteredOriginalData) &&
+            $targetContent instanceof EntityAbstract &&
```
vsolovei-smartling, line 80: This code relied on direct changing of class properties, this should not be needed

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -254,7 +250,11 @@ public function applyXML(SubmissionEntity $submission, string $xml, XmlHelper $x
                 $this->getLogger()->critical('Translation is not array after applying filter ' . ExportedAPI::FILTER_BEFORE_TRANSLATION_APPLIED . '. This is most likely due to an error outside of the plugins code.');
             }
             $translation = $this->externalContentManager->setExternalContent($original, $translation, $submission);
-            $this->setValues($targetContent, $translation['entity'] ?? []);
+            if ($targetContent instanceof EntityAbstract) {
+                $this->setValues($targetContent, $translation['entity'] ?? []);
```
vsolovei-smartling, line 254: Once again, direct manipulation of class properties, vs recreating the entity from relevant fields

inc/Smartling/ContentTypes/ContentTypeAbstract.php
```diff
@@ -2,86 +2,58 @@
```
vsolovei-smartling, line 2: I would like to further refactor this and it's descendants

inc/Smartling/ContentTypes/ContentTypeInterface.php
```diff
@@ -2,92 +2,41 @@
```
vsolovei-smartling, line 2: I've removed some methods from the interface, I don't want new entities to use ContainerBuilder to be constructed, and the removed methods were used in only some of the existing entities (registerIOWrapper, registerWidgetHandler, registerFilters)

inc/Smartling/ContentTypes/ContentTypeManager.php
```diff
@@ -62,51 +45,37 @@ public function getDescriptorsByBaseType($baseType)
...
         return $output;
     }
 
-    public function getHandler(string $contentType)
+    public function getHandler(string $contentType): ContentTypeInterface
     {
-        if (array_key_exists($contentType, $this->getCollection())) {
-            return $this->getCollection()[$contentType];
-        } else {
-            if (true === $this->getAllowDefault() && null !== $this->getDefaultHandler()) {
```
vsolovei-smartling, line 104: always false

inc/Smartling/ContentTypes/ContentTypePluggableInterface.php
```diff
@@ -6,10 +6,12 @@
 
 interface ContentTypePluggableInterface
 {
-    public function canHandle(string $contentType, int $contentId): bool;
+    public function canHandle(string $contentType, ?int $contentId = null): bool;
```
vsolovei-smartling, line 9: Changed to null to allow a handler to register custom post type when using e. g. Bulk Submit screen, no content ids there

inc/Smartling/ContentTypes/ContentTypePluggableInterface.php
```diff
@@ -6,10 +6,12 @@
...
 
     public function getContentFields(SubmissionEntity $submission, bool $raw): array;
 
+    public function getExternalContentTypes(): array;
```
vsolovei-smartling, line 13: Used in Bulk Submit, previously, all content was somehow linked to post-based content

inc/Smartling/ContentTypes/ContentTypeWidget.php
```diff
@@ -58,25 +44,11 @@ public function getSystemName()
      * Display name of content type, e.g.: Post
      * @return string
      */
-    public function getLabel()
+    public function getLabel(): string
     {
         return __('Theme Widget');
     }
 
-    /**
```
vsolovei-smartling, line 66: Default case of visible everywhere pushed to ContentTypeAbstract

inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
```diff
@@ -92,7 +92,6 @@ private function getAttachmentIds(array $data): array
 
     public function getContentFields(SubmissionEntity $submission, bool $raw): array
     {
-        $submission->assertHasSource();
```
vsolovei-smartling, line 95: Moved to calling code

inc/Smartling/ContentTypes/ExternalContentGravityForms.php
```diff
@@ -0,0 +1,173 @@
...
+    {
+        /*$fields = $this->siteHelper->withBlog($submission->getSourceBlogId(), function () use ($submission) {
+            return [
+                'entity' => $this->handler->getFormData($submission->getSourceId()),
+                'meta' => $this->handler->getMeta($submission->getSourceId()),
+            ];
+        });*/
+        $form = $this->handler->getFormData($submission->getSourceId());
+        $meta = $this->handler->getMeta($submission->getSourceId());
+        $displayMeta = json_decode($meta['display_meta'], true, 512, JSON_THROW_ON_ERROR);
```
vsolovei-smartling, line 73: Will refactor soon

inc/Smartling/ContentTypes/ExternalContentGravityForms.php
```diff
@@ -0,0 +1,171 @@
...
+
+    public function getMaxVersion(): string
+    {
+        return '2.5.9.3';
+    }
+
+    public function getMinVersion(): string
+    {
+        return '2.5.9.3';
+    }
```
PavelLoparev, line 130: So we support exact version but not a range of minor versions?

inc/Smartling/DbAl/WordpressContentEntities/GravityFormsFormHandler.php
```diff
@@ -0,0 +1,111 @@
...
+
+        return array_map(static function (array $item) {
+            return new GravityFormsForm('', $item['id'], $item['title'], $item['date_updated']);
+        }, $this->db->fetch($query, ARRAY_A));
+    }
+
+    #[ArrayShape([
+        'title' => 'string',
+        'date_updated' => 'string',
+    ])]
```
PavelLoparev, line 51: Every time we write this note it seems like we just need to introduce aninterface

readme.txt
```diff
@@ -4,7 +4,7 @@ Tags: translation, localization, localisation, translate, multilingual, smartlin
 Requires at least: 5.5
 Tested up to: 6.1
 Requires PHP: 8.0
-Stable tag: 3.0.4
+Stable tag: 3.1.0
```
PavelLoparev, line 7: I'm lost in these refactoring changes. Did we break any backward compatibility for developers? A lot of interfaces/classes changes.

inc/Smartling/Base/SmartlingCoreAttachments.php
PavelLoparev: I think we need to check if this is true for the older WP versions (minimal version we support). It might be the case in older major version it returned wp_error but not array

inc/Smartling/Base/SmartlingCore.php
```diff
@@ -70,7 +71,7 @@ public function prepareRelatedSubmissions(SubmissionEntity $submission)
             $submission->getId(),
         ]));
         $originalEntity = $this->getContentHelper()->readSourceContent($submission);
-        $relatedContentTypes = $originalEntity->getRelatedTypes();
+        $relatedContentTypes = $originalEntity instanceof EntityAbstract ? $originalEntity->getRelatedTypes() : [];
```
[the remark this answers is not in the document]
PavelLoparev: There are so many such conditions throughout the code. Maybe introduce some class with default methods that return needed default data (like empty array as needed here)?

inc/Smartling/Base/SmartlingCoreTrait.php
PavelLoparev: The same comment here. I mean some classes that extend entity abstract can implement this method with some meaningful actions and some just leave empty. So in this case you would not need to modify the "core" general code for all the entity types.

inc/Smartling/Base/SmartlingCore.php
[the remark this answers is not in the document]
vsolovei-smartling: Done

inc/Smartling/Base/SmartlingCoreAttachments.php
vsolovei-smartling: It's array in 5.5, which is the minimum we support

inc/Smartling/Base/SmartlingCoreTrait.php
vsolovei-smartling: EntityAbstract works this way, new entities should have no extra fields to clean up from the start, so I don't want to add this method to Entity, and EntityAbstract already has this method.

PavelLoparev (approved):

PR #450, opened 2023-03-17
fix Elementor translations not applied when custom components with non-string ids are present (WP-793)

vsolovei-smartling:


inc/Smartling/ContentTypes/ExternalContentElementor.php
```diff
@@ -166,15 +166,15 @@ private function extractContent(array $data, string $previousPrefix = ''): Exter
             }
             if (isset($component['settings'])) {
                 foreach ($component['settings'] as $key => $setting) {
-                    if (strpos($key, '_') === 0) {
+                    if (str_starts_with($key, '_')) {
```
PavelLoparev, line 169: This is available since php8. Did we drop support for php7?

inc/Smartling/ContentTypes/ExternalContentElementor.php
```diff
@@ -261,20 +261,24 @@ private function mergeElementorData(array $original, array $translation, Submiss
...
-                    if (strpos($settingIndex, '_') === 0) {
+                    if (str_starts_with($settingIndex, '_')) {
                         continue;
                     }
                     if (is_array($setting)) {
                         if (array_key_exists('id', $setting) && array_key_exists('url', $setting)) {
-                            $targetAttachmentId = $this->getTargetId($submission->getSourceBlogId(), $setting['id'], $submission->getTargetBlogId());
+                            try {
+                                $targetAttachmentId = $this->getTargetId($submission->getSourceBlogId(), $setting['id'], $submission->getTargetBlogId());
+                            } catch (\TypeError) {
```
PavelLoparev, line 271: Catch TypeError? What is the case?

inc/Smartling/ContentTypes/ExternalContentElementor.php
```diff
@@ -261,20 +261,24 @@ private function mergeElementorData(array $original, array $translation, Submiss
             }
             if (array_key_exists('settings', $component)) {
                 foreach($component['settings'] as $settingIndex => $setting) {
-                    if (strpos($settingIndex, '_') === 0) {
+                    if (str_starts_with($settingIndex, '_')) {
                         continue;
                     }
                     if (is_array($setting)) {
                         if (array_key_exists('id', $setting) && array_key_exists('url', $setting)) {
```
dimitrystd, line 268: I think better to verify type of `$setting['id']` value rather than add `catch`

inc/Smartling/ContentTypes/ExternalContentElementor.php
vsolovei-smartling: When a custom component has string id, it causes $this->getTargetId() to throw a TypeError, as it expects an integer

PavelLoparev: got it. then yes, I believe just follow Dmitry's comment

inc/Smartling/ContentTypes/ExternalContentElementor.php
vsolovei-smartling: Done

PR #466, opened 2023-06-27
add integration tests for beaver builder and elementor (WP-810)

vsolovei-smartling:


inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
```diff
@@ -110,9 +110,12 @@ public function getPluginId(): string
         return 'beaver_builder';
     }
 
-    public function getPluginPath(): string
+    public function getPluginPaths(): array
     {
-        return 'bb-plugin/fl-builder.php';
+        return [
```
vsolovei-smartling, line 115: When installing from wp cli, the folder seems to differ from the ui installation

inc/Smartling/ContentTypes/ExternalContentGravityForms.php
```diff
@@ -180,6 +180,9 @@ public function getRelatedContent(string $contentType, int $contentId): array
 
     public function setContentFields(array $original, array $translation, SubmissionEntity $submission): ?array
     {
+        if (!array_key_exists($this->getPluginId(), $translation)) {
```
vsolovei-smartling, line 183: A fix, handler reverts translation changes for other handlers in some cases

tests/IntegrationTests/tests/RelationsTest.php
```diff
@@ -248,7 +248,7 @@ private function verify(int $postId, int $cloneBlogId, int $translationBlogId, i
         foreach ($relations->getMissingReferences()[$cloneBlogId][ContentTypeHelper::POST_TYPE_ATTACHMENT] as $imageId) {
             $submission = $this->uploadDownload($this->createSubmissionForCloning(ContentTypeHelper::POST_TYPE_ATTACHMENT, $imageId));
             $this->assertEquals(SubmissionEntity::SUBMISSION_STATUS_COMPLETED, $submission->getStatus());
-            $this->assertNotEquals($submission->getSourceId(), $submission->getTargetId());
+            $this->assertNotEquals(0, $submission->getTargetId());
```
vsolovei-smartling, line 251: This restriction causes flaky tests due to submission creations in other tests

inc/Smartling/ContentTypes/ContentTypePluggableInterface.php
```diff
@@ -18,7 +18,7 @@ public function getMinVersion(): string;
 
     public function getPluginId(): string;
 
-    public function getPluginPath(): string;
+    public function getPluginPaths(): array;
```
dimitrystd, line 21: Probably add a comment here that it's a workaround, and in general, the plugin has one path. Something like
```
Returns an array with a single element containing the plugin's relative path.
In some cases the path can be resolved differently, then the plugin may include additional variations.
```

inc/Smartling/ContentTypes/ExternalContentAbstract.php
```diff
@@ -25,9 +25,11 @@ public function __construct(PluginHelper $pluginHelper, SubmissionManager $submi
...
     {
         $plugins = $this->wpProxy->get_plugins();
-        if (array_key_exists($this->getPluginPath(), $plugins)) {
-            return $this->wpProxy->is_plugin_active($this->getPluginPath()) &&
-                $this->pluginHelper->versionInRange($plugins[$this->getPluginPath()]['Version'] ?? '0', $this->getMinVersion(), $this->getMaxVersion());
+        foreach ($this->getPluginPaths() as $path) {
+            if (array_key_exists($path, $plugins) && $this->wpProxy->is_plugin_active($path) &&
+                $this->pluginHelper->versionInRange($plugins[$path]['Version'] ?? 0, $this->getMinVersion(), $this->getMaxVersion())) {
+                return true;
+            }
```
PavelLoparev, line 32: If `getPluginPaths()` returns multiple paths then should we check all of them for supported version or checking only one is sufficient?

inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
```diff
@@ -110,9 +110,12 @@ public function getPluginId(): string
...
+            'bb-plugin/fl-builder.php',
+            'beaver-builder-lite-version/fl-builder.php',
```
PavelLoparev, line 117: Is it possible these can be of different versions? If yes then `canHandle()` above must be modified to check all these paths for versions.

inc/Smartling/ContentTypes/ExternalContentGravityForms.php
PavelLoparev: Could you add test for this case?

inc/Smartling/ContentTypes/ExternalContentAbstract.php
vsolovei-smartling: We only need to check one

inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
vsolovei-smartling: canHandle() does that, processes a list of paths until it finds an active supported version, or returns false

inc/Smartling/ContentTypes/ExternalContentGravityForms.php
vsolovei-smartling: Not sure, what would such a test case check for?

PavelLoparev: > A fix, handler reverts translation changes for other handlers IN SOME CASES

inc/Smartling/ContentTypes/ExternalContentBeaverBuilder.php
PavelLoparev: Are these `bb-plugin` and `beaver-builder-lite-version` the same ones?

PavelLoparev: What if we support first but do not support second? In your case it will return true anyway

vsolovei-smartling: They are for the same plugin, the destination is different depending on the installation method

PavelLoparev (approved):

inc/Smartling/ContentTypes/ExternalContentGravityForms.php
vsolovei-smartling: `some cases` is confusing and vague, I'll provide more info for clarity:
When multiple handlers can handle translation (based on active supported plugins and post type), if Gravity Forms handler was not the first to set translation fields, but no Gravity Forms translation content was present (plugin active, but a post doesn't use any forms, for example), it erroneously returned `$original`, thus reversing any changes to `$translation`. ExternalContentManager expects handlers to return null when no changes are needed, this was my mistake which I didn't encounter previously when testing, due to using only one external plugin when developing.

PR #469, opened 2023-07-05
don't use XML when cloning (WP-813)

vsolovei-smartling:


inc/Smartling/Base/SmartlingCoreTrait.php
```diff
@@ -41,11 +41,11 @@ public function prepareTargetContent(SubmissionEntity $submission): SubmissionEn
      * Prepares a duplicate of source content for target site and links them.
      * To be used JUST BEFORE SENDING to Smartling
      */
-    protected function prepareTargetEntity(SubmissionEntity $submission, bool $forceClone = false): SubmissionEntity
+    protected function prepareTargetEntity(SubmissionEntity $submission): SubmissionEntity
```
vsolovei-smartling, line 44: `$forceClone` was never passed to function

inc/Smartling/Base/SmartlingCoreTrait.php
```diff
@@ -105,26 +99,15 @@ protected function prepareTargetEntity(SubmissionEntity $submission, bool $force
...
             unset ($metaValue);
             $this->getContentHelper()->writeTargetMetadata($submission, $metaFields);
         }
 
-        if (false === $update) {
-            $submission->setTargetId($targetContent->getId());
-            $submission = $this->getSubmissionManager()->storeEntity($submission);
-        }
-
         try {
```
vsolovei-smartling, line 128: Already set target id and saved earlier on lines 84-85

inc/Smartling/Jobs/UploadJob.php
```diff
@@ -60,7 +60,6 @@ private function processUploadQueue(): void
     private function processCloning(): void
     {
         while (($submission = $this->submissionManager->findSubmissionForCloning()) !== null) {
-            do_action(ExportedAPI::ACTION_SMARTLING_PREPARE_SUBMISSION_UPLOAD, $submission);
```
vsolovei-smartling, line 63: Happens inside ACTION_SMARTLING_CLONE_CONTENT anyway

inc/Smartling/Base/SmartlingCoreTrait.php
```diff
@@ -63,34 +63,28 @@ protected function prepareTargetEntity(SubmissionEntity $submission, bool $force
...
-             * @var array $_entity
-             */
-            foreach ($_entity as $k => $v) {
-                // Value is of `mixed` type
+            foreach ($filteredData['entity'] as $k => $v) {
                 $targetContent->{$k} = apply_filters(ExportedAPI::FILTER_SMARTLING_METADATA_FIELD_PROCESS, $k, $v, $submission);
             }
         }
 
-        if (false === $forceClone && $targetContent instanceof EntityWithPostStatus) {
```
PavelLoparev, line 86: Shouldn't this condition `!$submission->isCloned()` be included here instead of `false === $forceClone` as it was done above?

inc/Smartling/Jobs/UploadJob.php
PavelLoparev: Sounds like separate events to me

inc/Smartling/Base/SmartlingCoreTrait.php
vsolovei-smartling: No, because this code is used for translation as well as cloning, this should case relevant content types to change post status to draft when the target content is being created, to avoid untranslated content appearing on target sites. The above change allows to simply overwrite a submission target content when re-cloning a submission.

inc/Smartling/Jobs/UploadJob.php
vsolovei-smartling: The events stay separate, but there's no sense in making two calls to achieve one goal, and cloning does no upload anyway

dimitrystd (approved):

PavelLoparev (approved):

PR #471, opened 2023-07-12
remove batch uid when changing submission status to In Progress (WP-807)

vsolovei-smartling:
this currently does not happen when uploading files to a single locale

### DIFF
--- a/inc/Smartling/Base/SmartlingCoreUploadTrait.php
+++ b/inc/Smartling/Base/SmartlingCoreUploadTrait.php
@@ -142,7 +142,6 @@ trait SmartlingCoreUploadTrait
                     ]
                 );
                 $this->getLogger()->warning($message);
-                $submission->setBatchUid('');
                 $submission = $this->getSubmissionManager()
                     ->setErrorMessage($submission, 'There is no original content for translation.');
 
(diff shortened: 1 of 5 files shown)


inc/Smartling/Submissions/SubmissionEntity.php
```diff
@@ -315,6 +315,7 @@ public function setStatus(string $status): SubmissionEntity
                 $this->setCompletedStringCount(0);
                 break;
             case static::SUBMISSION_STATUS_IN_PROGRESS:
+                $this->setBatchUid('');
```
dimitrystd, line 318: We also discussed that Failed upload also must reset `batchUid`. Since we use DB table as message queue, we should emulate queue rules. When a message leaves the queue its data is removed. The submission upload flow is completed when it's successfully uploaded (In Progress) or upload fails (Failed).

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -142,7 +142,6 @@ public function getXMLFiltered(SubmissionEntity $submission): string
                     ]
                 );
                 $this->getLogger()->warning($message);
-                $submission->setBatchUid('');
```
PavelLoparev, line 145: Wait, this is some special case? Here we don't call `setStatus()` where you now reset `batchUid`, but still we reset batch uid here explicitly.

vsolovei-smartling: setErrorMessage() calls setSubmissionStatus('FAILED')

PavelLoparev (approved):

PR #472, opened 2023-07-18
add profile changes on blog removal, clean up profiles on bootstrap (WP-816)

vsolovei-smartling:
The cause of increased warnings count is not a loop. On bootstrap, the plugin tries to get a profile list, for which it generates labels based on blog settings. If a blog does not exist, a warning is generated.

### DIFF
--- a/inc/Smartling/Services/BlogRemovalHandler.php
+++ b/inc/Smartling/Services/BlogRemovalHandler.php
@@ -10,6 +10,7 @@ use Smartling\Submissions\SubmissionEntity;
 use Smartling\Submissions\SubmissionManager;
 use Smartling\WP\WPHookInterface;
 
+/** @noinspection PhpUnused */
 class BlogRemovalHandler implements WPHookInterface
 {
     use LoggerSafeTrait;
--- a/inc/Smartling/Services/BlogRemovalHandler.php
+++ b/inc/Smartling/Services/BlogRemovalHandler.php
@@ -27,7 +27,7 @@ class BlogRemovalHandler implements WPHookInterface
     public function register(): void
     {
         add_action('delete_blog', [$this, 'blogRemovalHandler']);
-        add_action('wp_delete_site', [$this, 'blogRemovalHandler51']);
+        add_action('wp_delete_site', [$this, 'siteRemovalHandler']);
     }
 
     /**
@@ -84,7 +84,7 @@ class BlogRemovalHandler implements WPHookInterface
         }
     }
 
-    public function blogRemovalHandler51(\WP_Site $site): void
+    public function siteRemovalHandler(\WP_Site $site): void
     {
         $this->blogRemovalHandler($site->blog_id);
     }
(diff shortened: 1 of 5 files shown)


inc/Smartling/Services/BlogRemovalHandler.php
```diff
@@ -3,118 +3,38 @@
...
+        private SettingsManager $settingsManager,
+        private SubmissionManager $submissionManager,
+    ) {}
 
     /**
      * Registers wp hook handlers. Invoked by wordpress.
      */
     public function register(): void
     {
         add_action('delete_blog', [$this, 'blogRemovalHandler']);
```
vsolovei-smartling, line 28: This hook is deprecated as of WP5.1

inc/Smartling/Services/BlogRemovalHandler.php
```diff
@@ -3,118 +3,38 @@
...
+        add_action('wp_delete_site', [$this, 'blogRemovalHandler51']);
```
vsolovei-smartling, line 29: This is the new hook, for our purposes

dimitrystd (approved):

inc/config/register-on-startup.yml
```diff
@@ -23,7 +23,7 @@ services:
       - [ "addService", [ "@helper.shortcode" ]]
       - [ "addService", [ "@helper.gutenberg" ]]
       - [ "addService", [ "@meta-field.processor.manager" ]]
-      - [ "addService", [ "@service.side-removal-handler"]]
+      - [ "addService", [ "@service.blog-removal-handler"]]
```
dimitrystd, line 26: It seems the `blog` is the deprecated word, otherwise why they call the new hook `wp_delete_site`.

dimitrystd (approved):

inc/Smartling/Services/BlogRemovalHandler.php
```diff
@@ -132,28 +53,49 @@ public function blogRemovalHandler($blogId)
...
+                    $this->apiWrapper->deleteFile($submission);
+                }
+            }
+        }
+
+        foreach ($this->settingsManager->getEntities() as $profile) {
+            if ($profile->getOriginalBlogId()->getBlogId() === $blogId) {
+                $this->settingsManager->deleteProfile($profile->getId());
+                $this->getLogger()->notice("Deleted profile profileId={$profile->getId()} while deleting blogId=$blogId");
+            } else {
```
PavelLoparev, line 73: Is it really ok to remove profile on blog removal? Sounds strange to me.

PavelLoparev (approved):

inc/Smartling/Services/BlogRemovalHandler.php
vsolovei-smartling: Not only leaving the profile would generate warnings, but since a deleted blog's id is not reused under normal circumstances, the profile would be unusable anyway, as it's source blog does not exist anymore

PR #473, opened 2023-07-27
use interval to add lock attributes (WP-808)

vsolovei-smartling:
this should satisfy both cases when attributes should be added immediately and after DOM load

### DIFF
--- a/js/smartling-connector-gutenberg-lock-attributes.js
+++ b/js/smartling-connector-gutenberg-lock-attributes.js
@@ -20,10 +20,8 @@ const addSmartlingGutenbergLockAttributes = function () {
                 return attributes;
             });
         }
-        return true;
     } else {
-        console.log('Smartling WordPress connector plugin is unable to add filter to wordpress hooks: no wp.hooks.addFilter.');
-        return false;
+        console.error('Smartling WordPress connector plugin is unable to add filter to wordpress hooks: no wp.hooks.addFilter.');
     }
 }
 
@@ -33,8 +31,6 @@ const generateSmartlingLockId = () => {
 
 const LOCK_ID_ATTRIBUTE = 'smartlingLockId';
 
-const interval = setInterval(() => {
-    if (addSmartlingGutenbergLockAttributes()) {
-        clearInterval(interval);
-    }
-}, 1000);
+document.addEventListener('DOMContentLoaded', function () {
+    addSmartlingGutenbergLockAttributes();
+});
--- a/js/smartling-connector-gutenberg-lock-attributes.js
+++ b/js/smartling-connector-gutenberg-lock-attributes.js
@@ -1,4 +1,9 @@
+let added = false;
+
 const addSmartlingGutenbergLockAttributes = function () {
+    if (added) {
+        return;
+    }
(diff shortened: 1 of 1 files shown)


js/smartling-connector-gutenberg-lock-attributes.js
```diff
@@ -31,6 +33,8 @@ const generateSmartlingLockId = () => {
...
 const LOCK_ID_ATTRIBUTE = 'smartlingLockId';
 
-document.addEventListener('DOMContentLoaded', function () {
-    addSmartlingGutenbergLockAttributes();
-});
+const interval = setInterval(() => {
+    if (addSmartlingGutenbergLockAttributes()) {
+        clearInterval(interval);
+    }
+}, 1000);
```
PavelLoparev, line 40: Maybe use `setTimeout()` for a single shot instead of `setInterval() + clearing timer?`

Also, it doesn't sound 100% reliable. You probably need to leave event listener for `DOMContentLoaded` event, page could load slower than 1 second.

PavelLoparev: Or probably we would need to check docs for Gutenberg-related events? As I get this you need to alter gutenberg blocks (and seems like they are loaded after `DOMContentLoaded` loaded) once they are loaded but again, they might load slower than 1 second.

Also, strtange that current approach worked for us but doesn't not for customer - does cusomer have customized FE code or custom theme or something? What is the root cause of changed sequents of events?

vsolovei-smartling: > Maybe use `setTimeout()` for a single shot instead of `setInterval() + clearing timer?`
> 
> Also, it doesn't sound 100% reliable. You probably need to leave event listener for `DOMContentLoaded` event, page could load slower than 1 second.

Exactly the reason for interval vs timer

PavelLoparev: > > Maybe use `setTimeout()` for a single shot instead of `setInterval() + clearing timer?`
> > Also, it doesn't sound 100% reliable. You probably need to leave event listener for `DOMContentLoaded` event, page could load slower than 1 second.
> 
> Exactly the reason for interval vs timer

Didn't get. You fire only once.

js/smartling-connector-gutenberg-lock-attributes.js
```diff
@@ -31,6 +37,15 @@ const generateSmartlingLockId = () => {
...
+
+if (!added) {
+    document.addEventListener('DOMContentLoaded', function () {
+        addSmartlingGutenbergLockAttributes();
+    });
+    const interval = setInterval(() => {
+        if (added) {
+            clearInterval(interval);
+        }
+    }, 1000);
```
PavelLoparev, line 53: Didn't you miss putting `addSmartlingGutenbergLockAttributes();` in the interval callback?

vsolovei-smartling: > Didn't you miss putting `addSmartlingGutenbergLockAttributes();` in the interval callback?

My bad, fixed!

TrilipuT (approved):

PR #478, opened 2023-08-11
add microseconds to log output (WP-822)

vsolovei-smartling:
Resulting log lines start with [2023-08-11 09:49:25.641179]

### DIFF
--- a/inc/Smartling/DITrait.php
+++ b/inc/Smartling/DITrait.php
@@ -3,14 +3,11 @@
 namespace Smartling;
 
 use Smartling\Base\ExportedAPI;
-use Smartling\Base\SmartlingLogHandler;
 use Smartling\Exception\SmartlingConfigException;
 use Smartling\Helpers\DiagnosticsHelper;
 use Smartling\Helpers\LogContextMixinHelper;
-use Smartling\MonologWrapper\Logger\LevelLogger;
 use Smartling\MonologWrapper\MonologWrapper;
 use Smartling\Services\GlobalSettingsManager;
-use Smartling\Vendor\Monolog\Handler\BufferHandler;
 use Smartling\Vendor\Monolog\Handler\NullHandler;
 use Smartling\Vendor\Symfony\Component\Config\FileLocator;
 use Smartling\Vendor\Symfony\Component\DependencyInjection\ContainerBuilder;
@@ -55,8 +52,6 @@ trait DITrait
         self::injectLoggerCustomizations(self::$containerInstance);
         self::handleLoggerConfiguration();
 
-        self::registerCloudLogExtension(self::$containerInstance);
-
         /**
          * Exposing reference to DI interface
          */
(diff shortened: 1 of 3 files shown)


inc/config/logger.yml
```diff
@@ -5,7 +5,7 @@ parameters:
   logger.filehandler.standard.loglevel: DEBUG
   logger.filehandler.standard.maxfiles: 1
 
-  logger.formatter.date: "Y-m-d H:i:s"
```
dimitrystd, line 8: Recommended format "yyyy-MM-dd'T'HH:mm:ss.SSSZ"

dimitrystd (changes requested):

inc/config/logger.yml
vsolovei-smartling: Now is [2023-08-11T13:49:24.079624Z]

PavelLoparev: Let's make sure Splunk correctly handles microseconds as well. lgtm

PavelLoparev (approved):

inc/Smartling/DITrait.php
```diff
@@ -81,7 +81,8 @@ protected static function initContainer()
     private static function registerCloudLogExtension(ContainerBuilder $di)
     {
         $defSmartling = $di->register('smartlingLogFileHandler', SmartlingLogHandler::class)
-            ->addArgument('https://api.smartling.com/updates/status');
+            ->addArgument('https://api.smartling.com/updates/status')
+            ->addArgument($di->getParameter('logger.formatter.date'));
```
vsolovei-smartling, line 85: Is there a reason not to move this to .yml entirely?

inc/config/logger.yml
vsolovei-smartling: Ok, some code changes are required
image

inc/Smartling/DITrait.php
dimitrystd: I don't know why. Probably because there are other logger variables `logger.formatter.*`. We can keep it there.

inc/config/logger.yml
dimitrystd: `u` - microseconds also should work, sitecore sends in this format. Can you pls check how it works in kibana? it must parse it, otherwise you will not see your logs.
`v` - milliseconds looks more common.

dimitrystd (approved):

PavelLoparev (approved):

vsolovei-smartling:
The kibana time stores only milliseconds, but I don't think we should rebuild the index to include microseconds. I've also refactored the code to be in the config, instead of DITrait

inc/config/logger.yml
```diff
@@ -56,4 +68,16 @@ services:
 
   logger:
     class: Smartling\MonologWrapper\Logger\LevelLogger
-    arguments: ["%logger.channel%", "%logger.filehandler.standard.loglevel%", ["@fileLoggerHandlerRotatable"]]
+    arguments:
+      - "%logger.channel%"
+      - "%logger.filehandler.standard.loglevel%"
+      -
+        - "@fileLoggerHandlerRotatable"
```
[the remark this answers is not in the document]
vsolovei-smartling: This is a nested array, the third parameter in LevelLogger expects an array, and we provide an array of one CustomRotatingFileHandler

PavelLoparev (approved):

dimitrystd (approved):

PR #482, opened 2023-09-04
add deliver translations endpoint (WP-827)

vsolovei-smartling:


PavelLoparev (commented):
Didn't we discuss that everything related to new rest api would be placed in the new branch and all PRs would be agains that new branch but not against master?

inc/Smartling/API/SettingsServiceApi.php
```diff
@@ -0,0 +1,25 @@
...
+    {
+        parent::__construct($projectId, self::initializeHttpClient(self::ENDPOINT_URL), $this->getLogger(), self::ENDPOINT_URL);
+        $this->setAuth($auth);
+    }
+
+    public function getSettings(): array
+    {
+        return $this->sendRequest('/integrations/wordpress/settings', $this->getDefaultRequestData('query', []), self::HTTP_METHOD_GET);
+    }
+}
```
PavelLoparev, line 25: Seems like it's time to update php sdk? But also, seems like it's time to split it to public and private parts?

inc/Smartling/API/SubmissionServiceApi.php
```diff
@@ -0,0 +1,127 @@
+<?php
+
+namespace Smartling\API;
+
+use Smartling\Models\AssetUid;
+use Smartling\Vendor\Smartling\AuthApi\AuthApiInterface;
+use Smartling\Vendor\Smartling\BaseApiAbstract;
+
+class SubmissionServiceApi extends BaseApiAbstract {
```
PavelLoparev, line 9: It's already in the php sdk. I think you need to use service from sdk and wordpress related code should be placed in api wrapper.

inc/Smartling/ApiWrapperWithRetries.php
```diff
@@ -60,6 +61,13 @@ public function downloadFile(SubmissionEntity $entity): string
         });
     }
 
+    public function getSettings(string $projectUid): array
+    {
+        return $this->withRetry(function () use ($projectUid) {
```
PavelLoparev, line 66: is it possible to write `withRetry` as an annotation?

inc/Smartling/Models/Settings.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function __construct(
+        private string $userIdentifier,
+        private string $secretKey,
+        private string $retrievalType,
+        private string $sourceLocale,
+        private int $sourceBlogId,
+        private array $targetLocales,
+    ) {
+    }
```
PavelLoparev, line 15: Is this the only settings we would like to store in settings service? What about all the regexps and others?

vsolovei-smartling:
> Didn't we discuss that everything related to the new rest API would be placed in the new branch and all PRs would be against that new branch but not against the master?

Good point, thanks, fixed

inc/Smartling/API/SettingsServiceApi.php
vsolovei-smartling: Yes, I don't yet want to update the SDKs until we agree on what should go where

inc/Smartling/ApiWrapperWithRetries.php
vsolovei-smartling: Yes, it should be, created https://bt.smartling.net/browse/WP-828 to research further and/or implement

dimitrystd (commented):
Didn't finish review

PR #483, opened 2023-09-06
RFC add retry attribute (WP-828)

vsolovei-smartling:
I've expected more from attributes. I'm kind of on the fence regarding this one.
Looking back at our previous conversations regarding retries (https://github.com/Smartling/wordpress-localization-plugin/pull/342#discussion_r754543711, https://github.com/Smartling/wordpress-localization-plugin/pull/355#discussion_r777719283) I feel like I was not explicit enough.
The choice of creating `ApiWrapperWithRetries` is due to the reason that we can't really proxy requests from one class to another. We have magic [__call](https://www.php.net/manual/en/language.oop5.overloading.php#object.call) that is triggered when invoking inaccessible methods in an object context, but our methods are accessible, they're public. Making them private and relying on `__call` would open another can of worms, I really dislike it as an option, the [comment](https://www.php.net/manual/en/language.oop5.overloading.php#119407) from 7 years old still stands strong. [__invoke](https://www.php.net/manual/en/language.oop5.overloading.php#object.invoke) is called when a script tries to call an object as a function e.g. `(new ApiWrapper('Hello')`, so not applicable.
Then come the attributes themselves. Having `#[Retry]` before a method call for a programmer coming from Java would look like, oh, hey, great, this will retry. It will not. [Attribute](https://www.php.net/manual/en/language.attributes.syntax.php) names and their arguments are resolved to a class and the arguments are passed to its constructor, when an instance of the attribute is requested through the Reflection API. So we still need some underlying class to work the attributes.
P. S.: No, it's not possible to add an attribute to an interface.
P. P. S.: On `ApiWrapperWithRetries implements ApiWrapperInterface` vs `ApiWrapperWithRetries extends ApiWrapper` implements was chosen due to the fact that having `#[Retry]` before a method could seem that the retry will happen. In the implements approach, a developer would be forced to update `ApiWrapperWithRetries`, if it would extend `ApiWrapper`, the margin of error is greater, as a developer would most likely update only `ApiWrapper` and then some time later we would discover that no retries are in fact happening.

### DIFF
--- a/inc/Smartling/ApiWrapper.php
+++ b/inc/Smartling/ApiWrapper.php
@@ -97,24 +97,28 @@ class ApiWrapper implements ApiWrapperInterface
         return $authProvider;
     }
 
+    #[Retry]
     public function acquireLock(ConfigurationProfileEntity $profile, string $key, int $ttlSeconds): \DateTime
     {
         return DistributedLockServiceApi::create($this->getAuthProvider($profile), $profile->getProjectId(), new NullLogger())
             ->acquireLock("{$profile->getProjectId()}-$key", $ttlSeconds);
     }
 
+    #[Retry]
     public function renewLock(ConfigurationProfileEntity $profile, string $key, int $ttlSeconds): \DateTime
     {
         return DistributedLockServiceApi::create($this->getAuthProvider($profile), $profile->getProjectId(), $this->getLogger())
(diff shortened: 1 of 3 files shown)


vsolovei-smartling:
Decided against merging

PR #484, opened 2023-09-12
add check for submission source uniqueness, add UI to resolve multiple target submissions for one source (WP-829)

vsolovei-smartling:
No constraints are present in the database to enforce uniqueness, this causes duplicate submissions referencing single asset multiple times in the source blog in cases of complex relations.
Adding such a constraint right now could fail for already affected users. I've created a new menu page that would only show if the issue is present, and will roll out the db migration later.


vsolovei-smartling:
New UI looks like this:
image

inc/Smartling/Submissions/SubmissionManager.php
```diff
@@ -431,6 +429,10 @@ public function storeEntity(SubmissionEntity $entity): SubmissionEntity
         $tableName = $this->getDbal()->completeTableName(SubmissionEntity::getTableName());
 
         if ($is_insert) {
+            if ($this->submissionExists($entity->getContentType(), $entity->getSourceBlogId(), $entity->getSourceId(), $entity->getTargetBlogId())) {
+                $this->getLogger()->info("Skip inserting submission=$originalSubmission, submission for same source and target already exists");
+                return $this->getSubmissionEntity($entity->getContentType(), $entity->getSourceBlogId(), $entity->getSourceId(), $entity->getTargetBlogId());
```
dimitrystd, line 434: This line returns data from DB, but you lose some new values from `$fields`. I think your code should fallback to "update" flow.

vsolovei-smartling: I have to disagree. If a proper constraint would be present on the database, we would get an exception while trying to insert a new submission for same content.

PavelLoparev (approved):

PR #485, opened 2023-09-13
add cron throtling (WP-830)

vsolovei-smartling:


inc/Smartling/Jobs/JobAbstract.php
```diff
@@ -182,6 +181,11 @@ private function tryRunJob(): string
             $message = "Failed to place lock flag for {$this->getJobHookName()}: $errorMessage";
             $this->getLogger()->debug($message);
             return $message;
+        } catch (\RuntimeException $e) {
+            if ($e->getMessage() === self::THROTTLED_MESSAGE) {
+                return $e->getMessage();
+            }
+            throw $e;
```
PavelLoparev, line 188: `tryRunJob()` has strange behavior. In the success case, it returns an empty string. In an error case, it can throw or can return non-empty string. Should be `return void or throw` instead. Why caller code can't catch exception and check error message on its own? Or we already have dozen if/else which relies on empty string on success case and non empty string in error case and sometimes or exceptions?

vsolovei-smartling: Moved up

inc/config/cron.yml
```diff
@@ -3,6 +3,7 @@ parameters:
   cron.interval.group.submission: "3h"
   cron.interval.last-modified-check: "5m"
   cron.interval.download: "5m"
+  cron.interval.throttle: 60
```
dimitrystd, line 6: IMO, once per minute is still a hight frequency. I would recommend at least 120.

dimitrystd (approved):

inc/config/plugin-info.yml
```diff
@@ -5,3 +5,7 @@ parameters:
   plugin.url: "@=service('wp.proxy').plugin_dir_url(constant('SMARTLING_PLUGIN_DIR')~constant('DIRECTORY_SEPARATOR')~'..')"
   plugin.version: "auto updated"
   plugin.domain: %plugin.name%
+
+#TODO no commit
```
dimitrystd, line 9: :)
Push with "force" to remove this commit from the history.

dimitrystd (changes requested):

inc/config/plugin-info.yml
vsolovei-smartling: :(

inc/config/cron.yml
vsolovei-smartling: Done

inc/config/plugin-info.yml
PavelLoparev: what was there? didn't notice those secrets)

PavelLoparev: lol
image

if you click on `from` commit [you will still see that secrets](https://github.com/Smartling/wordpress-localization-plugin/commit/1852398374e3572d10a312f88567e9ffc0addab8)

PavelLoparev: not sure why after force pushing it's still there. Did you do:
```
git reset HEAD~N // commit before your commit with secrets
// remove you secrets
git add ...
git commit ...
git push ...
```
?

dimitrystd (approved):

PavelLoparev (approved):

PR #486, opened 2023-09-14
add bulk submit screen filter api, remove broken status filtering (WP-831)

vsolovei-smartling:
WordPress doesn't allow to get posts order by status


inc/Smartling/WP/View/BulkSubmit.php
```diff
@@ -121,6 +124,4 @@
             });
         })(jQuery);
     </script>
-
-    </form>
```
vsolovei-smartling, line 125: Form already closed on line 112

PavelLoparev (approved):

PR #490, opened 2023-10-19
add support for Elementor icon list widget (WP-836)

vsolovei-smartling:


inc/Smartling/ContentTypes/ExternalContentElementor.php
```diff
@@ -259,7 +259,23 @@ private function getRelatedFromSetting(array $setting): ?array {
...
 
+        $ids = [];
+        foreach ($setting as $value) {
+            if (is_array($value)) {
+                $related = $this->getRelatedFromSetting($value);
+                if ($related !== null) {
+                    $ids[] = array_values($related)[0][0];
+                }
+            }
+        }
```
PavelLoparev, line 274: So, this recursion has 2 conditions to exit. If there are no `source` or `library` properties in `$settings` variable then it will return just null, right?

dimitrystd (approved):

inc/Smartling/ContentTypes/ExternalContentElementor.php
vsolovei-smartling: Yes

PavelLoparev (approved):

PR #491, opened 2023-11-02
add block level locking when cloning (WP-838)

vsolovei-smartling:


inc/Smartling/Base/SmartlingCore.php
```diff
@@ -58,11 +58,24 @@ public function __construct(ExternalContentManager $externalContentManager, Gute
...
+                return '';
+            });
+        }
         $submission = $this->prepareTargetContent($submission);
         $entity = $this->getContentHelper()->readTargetContent($submission);
         $content = $entity->toArray();
         foreach ($content as $key => $value) {
             if (is_string($value)) {
+                if ($key === 'post_content') {
+                    $value = $this->postContentHelper->applyTranslationsWithLockedBlocks($target, $value);
```
dimitrystd, line 77: Maybe you need to add a comment here. It's not clear what exactly going here.
The `$content` comes from `readTargetContent($submission)`. The `$target` comes from `get_post($submission->getTargetId())`. It's hard to guess what is inside, but it looks like they read exactly the same target post.

dimitrystd: It triggered my attention because the new method has the word `Translations`, but it's not obvious why do we need translation inside the "clone" method.

readme.txt
```diff
@@ -62,6 +62,9 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
+= 3.6.0 =
+* Added support for block level locking when cloning content
```
dimitrystd, line 66: Could you please elaborate on this by adding some details? What exactly was change compared to behavior before?

dimitrystd (approved):

PavelLoparev (approved):

PR #498, opened 2023-12-04
add check source block exists (WP-852) 

vsolovei-smartling:
Adding a nested Gutenberg block and locking it caused the connector to throw a fatal error while trying to download the content

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
   "name": "smartling/wordpress-connector",
   "license": "GPL-2.0-or-later",
-  "version": "3.8.0",
+  "version": "3.8.1",
   "description": "",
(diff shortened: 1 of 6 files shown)


vsolovei-smartling:
> Approved but I didn't get why not having some extra content in a source block may cause the issue with target block.

It's not a source block, it's extra content in a target block.
When the connector plugin applies translation, it overwrites source strings with translated strings, unless the content is locked. If it is locked, the plugin tries to apply a string from the target instead, but fails because it's unable to get the corresponding source string to alter.
That was a good, thought provoking question. In time we'll need to rebuild the way this works entirely to allow for proper 3-way merging due to pending feature requests.

PR #502, opened 2024-01-05
always remove known untranslatable fields if detected (WP-859)

vsolovei-smartling:


inc/Smartling/ContentTypes/ExternalContentManager.php
```diff
@@ -23,35 +29,26 @@ public function addHandler(ContentTypePluggableInterface $handler): void
...
-                    }
-                    if ($handler instanceof ContentTypeModifyingInterface) {
-                        try {
-                            $source = $handler->alterContentFieldsForUpload($source);
-                        } catch (\Throwable $e) {
-                            $this->getLogger()->notice('HandlerName="' . $handler->getPluginId() . '" got exception while trying to alter content fields: ' . $e->getMessage());
-                        }
-                    }
-                    break;
-                case ContentTypePluggableInterface::VERSION_NOT_SUPPORTED:
```
PavelLoparev, line 43: Why don't we need this case anymore?

vsolovei-smartling: The point is to always check and remove these fields, they never need to be sent, as they cause various issues, and the translation applied is not displayed if the plugins are active

PavelLoparev (approved):

PR #504, opened 2024-01-17
add bulk submit screen submission status filter (WP-864)

vsolovei-smartling:


inc/Smartling/WP/Table/BulkSubmitTableWidget.php
```diff
@@ -350,66 +297,81 @@ public function prepare_items()
...
-            ($pageOptions['page'] - 1) * $pageOptions['limit'],
-            $sortOptions['orderby'],
-            $sortOptions['order'],
-            $searchString
-        );
-
-
-        $total = $io->getTotal();
+        $contentIds = [];
+        if ($submissionStatusFilterValue !== null) {
```
vsolovei-smartling, line 309: The reason for added complexity is that if we provide the `include` parameter with the list of ids, WP changes the limit to include all supplied ids. If submission status filter is applied, we can no longer select all submission ids from the table to populate `include` parameter, but need to do paging ourselves.

inc/Smartling/WP/Table/BulkSubmitTableWidget.php
```diff
@@ -350,66 +297,81 @@ public function prepare_items()
...
+                foreach ($entities as $entity) {
+                    try {
+                        $blogLabel = $this->siteHelper
                             ->getBlogLabelById($this->localizationPluginProxy, $entity->getTargetBlogId());
+                    } catch (BlogNotFoundException) {
+                        $blogLabel = "UNKNOWN (blogId={$entity->getTargetBlogId()})";
                     }
-
-                    $row['locales'] = implode(', ', $locales);
+                    $locales[] = "<div style=\"display: inline-block\">$blogLabel
```
vsolovei-smartling, line 350: I've also added some colors:
image

inc/Smartling/DbAl/WordpressContentEntities/EntityAbstract.php
```diff
@@ -169,8 +169,15 @@ abstract public function get(mixed $id): self;
...
-    // It's not good idea to ask instence of Post\Tag class to return all objects
-    abstract public function getAll(int $limit = 0, int $offset = 0, string $orderBy = '', string $order = '', string $searchString = ''): array;
+    // It's not a good idea to ask an instance of Post\Tag class to return all objects
+    abstract public function getAll(
+        int $limit = 0,
+        int $offset = 0,
+        string $orderBy = '',
+        string $order = '',
+        string $searchString = '',
+        array $includeOnlyIds = [],
```
dimitrystd, line 179: I would recommend to call it `ids`. In java REST we keep filter names simple: `searchTerm`, `category`, `dateFrom`, etc. This new parameter also looks like a new search field.

inc/Smartling/WP/Table/BulkSubmitTableWidget.php
dimitrystd: Do you have example of SQL for `getAll()`? It may help to understand the problem. 
As far as I understand, we will have 2 SQL queries now.

vsolovei-smartling: The example SQL for posts looks like
`			SELECT   wp_posts.*
			FROM wp_posts 
			WHERE 1=1  AND wp_posts.post_type = 'post' AND ((wp_posts.post_status <> 'trash' AND wp_posts.post_status <> 'auto-draft'))
			
			ORDER BY wp_posts.post_date ASC
			LIMIT 0, 20`
it's generated by [get_posts()](https://developer.wordpress.org/reference/functions/get_posts/) WP function. If we are to provide an array of one id to the same call, the query becomes
`			SELECT   wp_posts.*
			FROM wp_posts 
			WHERE 1=1  AND wp_posts.ID IN (1) AND wp_posts.post_type = 'post' AND ((wp_posts.post_status <> 'trash' AND wp_posts.post_status <> 'auto-draft'))
			
			ORDER BY wp_posts.post_date ASC
			LIMIT 0, 1`
(note the limit has changed despite requested amount not being changed).
But this is only for posts, other content types have different methods of getting count of all, for example, for Gravity Forms support we query the database directly.
Can confirm, when Submission Status filter is used, we first query the submissions table to get relevant ids with paging, then provide them to get_posts()

inc/Smartling/DbAl/WordpressContentEntities/EntityAbstract.php
vsolovei-smartling: Renamed

inc/Smartling/Submissions/SubmissionEntity.php
```diff
@@ -350,7 +350,7 @@ public function getStatusColor(): string
             static::SUBMISSION_STATUS_FAILED => 'red',
         ];
 
-        return $statusColors[$this->getStatus()];
+        return $statusColors[$this->getStatus()] ?? 'red';
```
PavelLoparev, line 353: Maybe use `static::SUBMISSION_STATUS_FAILED` from above?

PavelLoparev (approved):

inc/Smartling/Submissions/SubmissionEntity.php
vsolovei-smartling: Good one, thanks

PR #494, opened 2023-11-15
add after target content written action (WP-841)

vsolovei-smartling:

PR #507, opened 2024-01-23
fix block attributes locking (WP-843)

vsolovei-smartling:


tests/IntegrationTests/tests/GutenbergBlocksTest.php
```diff
@@ -55,7 +132,6 @@ public function testInnerBlocks()
         $submissions[] = $this->translationHelper->prepareSubmission('post', $this->sourceBlogId, $postIds[3], $this->targetBlogId);
 
         foreach ($submissions as $submission) {
-            $submission->getFileUri();
```
PavelLoparev, line 58: Why did we need this before?

inc/Smartling/Helpers/PostContentHelper.php
```diff
@@ -159,7 +159,15 @@ private function getBlockByPath(array $blocks, string $path): ?GutenbergBlock
...
-        return $this->setBlock($blocks, $path, $gutenbergBlock)['block']->getInnerBlocks();
+        $result = $this->setBlock($blocks, $path, $gutenbergBlock);
+
+        if ($result['useInnerBlock']) {
+            return $result['block']->getInnerBlocks();
+        }
+
+        $blocks[$result['index']] = $result['block'];
+
+        return $blocks;
```
PavelLoparev, line 170: I didn't get the fix. It may return inner blocks or if there are no inner ones then you just add "root" block to blocks array?

tests/IntegrationTests/tests/GutenbergBlocksTest.php
vsolovei-smartling: [getFileUri used to have side effects](https://github.com/Smartling/wordpress-localization-plugin/pull/494/files#diff-734781b35b45fb97be3758e309ef129b85ca551d4376e945f1c5f69b87687fe5R451) which were relied on. This is no longer the case, the call is now pure.

inc/Smartling/Helpers/PostContentHelper.php
vsolovei-smartling: The function this one relies on returns replacement results and instructions how to handle it. Previously, a replaced block's that had locked attributes on root level inner blocks would be used instead of replacing a root level block with the replacement

PavelLoparev (approved):

dimitrystd (commented):
👀

PR #508, opened 2024-01-26
add impersonation (WP-865)

vsolovei-smartling:
The issue was with Elementor checking the permissions, and cron user has no permissions

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
   "name": "smartling/wordpress-connector",
   "license": "GPL-2.0-or-later",
-  "version": "3.8.5",
+  "version": "3.8.6",
   "description": "",
   "type": "wordpress-plugin",
   "repositories": [
--- a/inc/Smartling/Base/ExportedAPI.php
+++ b/inc/Smartling/Base/ExportedAPI.php
@@ -198,7 +198,7 @@ interface ExportedAPI
      */
     public const ACTION_AFTER_TARGET_CONTENT_WRITTEN = 'smartling_action_after_target_content_written';
 
-    public const ACTION_AFTER_TARGET_METADATA_WRITTEN = 'smartling_action_after_target_content_written';
+    public const ACTION_AFTER_TARGET_METADATA_WRITTEN = 'smartling_action_after_target_metadata_written';
 
     /**
      * @param SubmissionEntity
--- a/inc/Smartling/Base/ExportedAPI.php
+++ b/inc/Smartling/Base/ExportedAPI.php
@@ -198,6 +198,8 @@ interface ExportedAPI
      */
     public const ACTION_AFTER_TARGET_CONTENT_WRITTEN = 'smartling_action_after_target_content_written';
 
+    public const ACTION_AFTER_TARGET_METADATA_WRITTEN = 'smartling_action_after_target_content_written';
+
     /**
      * @param SubmissionEntity
      */
--- a/inc/Smartling/Helpers/UserHelper.php
+++ b/inc/Smartling/Helpers/UserHelper.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Smartling\Helpers;
+
+use Smartling\DbAl\SmartlingToCMSDatabaseAccessWrapperInterface;
+
+class UserHelper {
+    use LoggerSafeTrait;
+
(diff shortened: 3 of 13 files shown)


vsolovei-smartling:
Verified on their staging

composer.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "smartling/wordpress-connector",
   "license": "GPL-2.0-or-later",
-  "version": "3.8.5",
+  "version": "3.8.6",
```
PavelLoparev, line 4: Sounds like this is new non bc feature rather than bug fix. Shouldn't it be `3.9.0` instead?

inc/Smartling/Base/ExportedAPI.php
```diff
@@ -198,6 +198,8 @@ interface ExportedAPI
      */
     public const ACTION_AFTER_TARGET_CONTENT_WRITTEN = 'smartling_action_after_target_content_written';
 
+    public const ACTION_AFTER_TARGET_METADATA_WRITTEN = 'smartling_action_after_target_content_written';
```
PavelLoparev, line 201: Value is the same as in `ACTION_AFTER_TARGET_CONTENT_WRITTEN` constant.

inc/Smartling/Helpers/UserHelper.php
```diff
@@ -0,0 +1,55 @@
...
+            return $function();
+        }
+
+        $privilegedId = $this->getAdministratorOrEditorId();
+        $originalUserId = $this->wp->get_current_user_id();
+
+        if ($privilegedId === null) {
+            $this->getLogger()->warning("Unable to get administrator or editor for blogId=" . $this->wp->get_current_blog_id() . ', running as original userId=' . $originalUserId);
+            return $function();
+        }
```
PavelLoparev, line 33: Maybe throw here? Sounds like for the client's code it's not expected to proceed with non privileged user in `asAdministratorOrEditor` .

PavelLoparev (changes requested):

inc/Smartling/Base/ExportedAPI.php
vsolovei-smartling: Thanks, fixed!

inc/Smartling/Helpers/UserHelper.php
vsolovei-smartling: I had my doubts here, no huge difference for now, will throw

PavelLoparev: I mean, if it doesn't throw then probably client will face the same issue as before, right? I don't insist on throwing, just trying think outloud.

PavelLoparev (approved):

inc/Smartling/Helpers/UserHelper.php
```diff
@@ -28,8 +29,7 @@ public function asAdministratorOrEditor(callable $function): mixed
         $originalUserId = $this->wp->get_current_user_id();
 
         if ($privilegedId === null) {
-            $this->getLogger()->warning("Unable to get administrator or editor for blogId=" . $this->wp->get_current_blog_id() . ', running as original userId=' . $originalUserId);
-            return $function();
+            throw new SmartlingHumanReadableException('Unable to get administrator or editor to process content in blog ' . $this->wp->get_current_blog_id(), 'no.privileged.user', 403);
```
PavelLoparev, line 32: Oh, we have a similar approach in SUBM (and I'm not proud of that) where the business logic layer throws infrastructure layer exceptions (403 etc). Probably not part of this PR, but somewhen we would need to reconsider error handling. The service layer must throw service exceptions and the error handler on the infrastructure level must map those service layer exceptions to something other.

inc/Smartling/Helpers/UserHelper.php
vsolovei-smartling: Yes, it's better to throw instead of trying to continue as user without permissions, logs will be clearer

inc/Smartling/Helpers/UserHelper.php
vsolovei-smartling: Introduced new exception

dimitrystd (commented):
I believe you had a text file where you noted all major changes in "API". Just wanted to remind and confirm that we don't need it anymore since no 3rd party developers use our integration.

PR #509, opened 2024-01-31
add source blog absolute link rewrites (WP-869)

vsolovei-smartling:
Despite disliking adding settings, I think the new behaviour should be a toggle

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -4,7 +4,7 @@ Tags: translation, localization, localisation, translate, multilingual, smartlin
 Requires at least: 5.5
 Tested up to: 6.4.1
 Requires PHP: 8.0
-Stable tag: 3.8.6
+Stable tag: 3.9.0
 License: GPLv2 or later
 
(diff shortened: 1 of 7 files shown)


readme.txt
```diff
@@ -62,6 +62,9 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
+= 3.9.0 =
+* Changed default behaviour for absolute links that point to WordPress installation on the original site to point to the translated site instead. This could cause images that were previously loaded from the original site working on translated sites without being sent for translation to break in this version of plugin.
```
PavelLoparev, line 66: I don't get why we must implement some custom redirect logic applied on client's side in connector?

vsolovei-smartling: This definitely needs to be discussed, will do

vsolovei-smartling:
Will use #510 instead

PR #512, opened 2024-02-12
add support for yoast multiple key phrases and synonyms (WP-872)

vsolovei-smartling:


inc/Smartling/ContentTypes/ExternalContentGravityForms.php
```diff
@@ -66,6 +66,9 @@ public function getSupportLevel(string $contentType, ?int $contentId = null): st
 
     public function getContentFields(SubmissionEntity $submission, bool $raw): array
     {
+        if ($submission->getContentType() !== self::CONTENT_TYPE) {
```
vsolovei-smartling, line 69: Not task related, something I've noticed in the logs

inc/Smartling/ContentTypes/ExternalContentManager.php
```diff
@@ -21,7 +21,6 @@ public function __construct(
      */
     private array $handlers = [];
 
-    /** @noinspection PhpUnused, used in DI */
```
vsolovei-smartling, line 24: Also used in tests

inc/Smartling/ContentTypes/ExternalContentGravityForms.php
```diff
@@ -50,7 +50,7 @@ public function __construct(
         }
     }
 
-    public function removeUntranslatableFieldsForUpload(array $source): array
+    public function removeUntranslatableFieldsForUpload(array $source, SubmissionEntity $submission): array
```
PavelLoparev, line 53: We pass but we don't use?

PavelLoparev (changes requested):

inc/Smartling/ContentTypes/ExternalContentGravityForms.php
vsolovei-smartling: It is used only in the new Yoast handler so far, but I can see it being required in the future handlers as well

vsolovei-smartling: An alternative would be to do ?SubmissionEntity $submission = null in the interface, but it would require checking for provided Submission where it is needed, and I suspect there will be more checks than calls of the method in the code

PavelLoparev: my bad,makes sense.

PavelLoparev (approved):

inc/Smartling/ContentTypes/ExternalContentYoast.php
```diff
@@ -0,0 +1,119 @@
...
+use Smartling\Submissions\SubmissionEntity;
+use Smartling\Submissions\SubmissionManager;
+
+class ExternalContentYoast extends ExternalContentAbstract implements ContentTypeModifyingInterface {
+    use LoggerSafeTrait;
+
+    private array $handledFields = [
+        '_yoast_wpseo_focuskeywords',
+        '_yoast_wpseo_keywordsynonyms',
+    ];
```
PavelLoparev, line 19: Just thinking out loud: how can we add support for new fields without updating plugin code? Is it possible, for example, to introduce some "hook" that will allow developers to alter this default list of fields if there will be in the future?

dimitrystd: All existing hooks were added with this idea in mind (someone will add support for our integration). Wrong idea. We have to add by ourselves 🤬.

dimitrystd (approved):

PR #519, opened 2024-02-26
refactor (WP-874)

vsolovei-smartling:
Controversial, because on one side, I'm not a fan of the god object. On the other side, however, having traits that pretend they're usable without the root object makes it really harder to write tests and understand code in general.


inc/Smartling/Base/SmartlingCoreAttachments.php
```diff
@@ -93,58 +95,39 @@ public function regenerateTargetThumbnailsBySubmission(SubmissionEntity $submiss
             ])
         );
 
-        $this->getContentHelper()->ensureTargetBlogId($submission);
+        $this->siteHelper->withBlog($submission->getTargetBlogId(), function () use ($submission) {
```
PavelLoparev, line 98: It was `ensureTargetBlogId` now it's `withBlog`, is it expected change? I thought you've just got rid of traits and moved everything inside classes?

PavelLoparev (approved):
I'm afraid of such refactorings. But if all tests are passed then I'm ok

PR #520, opened 2024-02-27
use Smartling placeholders for shortcode detection (WP-879)

vsolovei-smartling:
The connector uses the shortcodes that are registered in WP or added in fine-tuning section. It appears the fine-tuning was not set up, as when testing locally with the proper setup, no `#sl-start#` or `#sl-end#` tags appear. Without setting up the fine-tuning, the result is as experienced by the reporter. I'll update the documentation accordingly, but this PR could also help to resolve the issue. The plugin only wraps registered shortcodes in said tags, and the potential for false-positives is quite low, I believe.

I could think of no other way to determine shortcodes presence, as a string that looks like a WP shortcode `[text]` might as well be plain text that is meant for translation.

### DIFF
--- a/inc/Smartling/Helpers/PlaceholderHelper.php
+++ b/inc/Smartling/Helpers/PlaceholderHelper.php
@@ -12,6 +12,11 @@ class PlaceholderHelper
         return self::SMARTLING_PLACEHOLDER_MASK_START . $string . self::SMARTLING_PLACEHOLDER_MASK_END;
     }
 
+    public function hasPlaceholders(string $string): bool
+    {
+        return preg_match('~' . self::SMARTLING_PLACEHOLDER_MASK_START . '.+' . self::SMARTLING_PLACEHOLDER_MASK_END . '~', $string);
+    }
+
     public function removePlaceholders(string $string): string
     {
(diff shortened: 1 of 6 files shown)


inc/Smartling/Helpers/PlaceholderHelper.php
```diff
@@ -12,6 +12,11 @@ public function addPlaceholders(string $string): string
         return self::SMARTLING_PLACEHOLDER_MASK_START . $string . self::SMARTLING_PLACEHOLDER_MASK_END;
     }
 
+    public function hasPlaceholders(string $string): bool
+    {
+        return preg_match('~' . self::SMARTLING_PLACEHOLDER_MASK_START . '.+' . self::SMARTLING_PLACEHOLDER_MASK_END . '~', $string);
```
dimitrystd, line 17: Use non non-greedy pattern. Test this string `#sl-start#test1#sl-end# test2 #sl-start#test3#sl-end#`
```suggestion
        return preg_match('~' . self::SMARTLING_PLACEHOLDER_MASK_START . '.+?' . self::SMARTLING_PLACEHOLDER_MASK_END . '~', $string);
```

vsolovei-smartling: We only detect placeholder strings without matching, I've added a test, and it passes as-is

PavelLoparev (approved):

dimitrystd (approved):

PR #521, opened 2024-03-01
switch to source blog when working with external content (WP-877)

vsolovei-smartling:


inc/Smartling/ContentTypes/ExternalContentGravityForms.php
```diff
@@ -58,7 +58,12 @@ public function removeUntranslatableFieldsForUpload(array $source, SubmissionEnt
 
     public function getSupportLevel(string $contentType, ?int $contentId = null): string
     {
-        if ($contentType === self::CONTENT_TYPE || $this->contentTypeHelper->isPost($contentType)) {
+        if ($contentType === self::CONTENT_TYPE
+            || ($this->contentTypeHelper->isPost($contentType)
+                && str_contains(
```
dimitrystd, line 63: Is it related to the original issue?
And want to confirm that the gravity plugin always puts the `gravityforms` word in the post so other posts will be ignored.

dimitrystd (approved):

inc/Smartling/ContentTypes/ExternalContentGravityForms.php
vsolovei-smartling: This is unrelated, but another issue I found while testing the external content. It's not critical, but causes some unnecessary logs. This second part of the check looks for the Gutenberg block that is named gravityforms-form. It's not critical if the gravityforms is present, but not a part of gravity forms content.

PR #523, opened 2024-03-21
add upload queue (WP-874)

vsolovei-smartling:


inc/Smartling/Base/SmartlingCore.php
```diff
@@ -147,7 +145,18 @@ protected function sendFile(SubmissionEntity $submission, $xmlFileContent, array
                 return false;
             }
 
-            $result = $this->getApiWrapper()->uploadContent($submission, '', $tmp_file, $smartlingLocaleList);
+            if ($jobUid === '') {
+                $jobUid = $this->getApiWrapper()->retrieveJobInfoForDailyBucketJob($profile)->getJobInformationEntity()->getJobUid();
+            }
+            $batchUid = $this->getApiWrapper()->createBatch($profile, $jobUid, $profile->getAutoAuthorize())['batchUid'];
```
PavelLoparev, line 151: It doesn't sounds right, we don;t need to create a separate batch for every file upload

inc/Smartling/DbAl/DB.php
```diff
@@ -134,26 +125,25 @@ private function installDb(): int
 
     public function schemaUpdate(int $fromVersion): void
     {
-        $prefix = $this->getWpdb()->base_prefix;
+        $prefix = $this->wpdb->base_prefix;
         foreach ($this->getMigrationManager()->getMigrations($fromVersion) as $migration) {
-            $this->logger->info('Starting applying migration ' . $migration->getVersion());
+            $this->getLogger()->info("Starting applying migration {$migration->getVersion()}");
```
PavelLoparev, line 130: `getWpdb()` -> `wpdb` but `logger` -> `getLogger()`. why?

PavelLoparev (changes requested):
comment for submitting reqview

inc/Smartling/DbAl/DB.php
vsolovei-smartling: `getLogger()` is from `LoggerSafeTrait`. Logger is set to null there, and a getLogger call allows to properly initialize logger only when it is needed.

PavelLoparev (approved):
Discussed with @vsolovei-smartling, we agreed to leave this as is but create a follow up to reconsider how we use batch api here. Several options:
1. Use Batch API v2 so that we don't think about when to execute the batch thus we don't need to create batch per file upload
2. Use Batch API v1 as now but pass additional info to the queue item like `executeAfterProcesingQueueItem<id>` so there will be no issue like `batch is already executed` etc (only if queue is being processed sequentially).

inc/Smartling/Base/SmartlingCore.php
vsolovei-smartling: Discussed, created a placeholder https://bt.smartling.net/browse/WP-881

PR #525, opened 2024-04-08
add upload queue (WP-881)

vsolovei-smartling:


PavelLoparev (commented):
I'm confused, I thought switching to batch api v2 it's only about introducing new sdk dependency + rewriting upload queue handler/worker. Why are there so many changes to core code (massive refactoring of core trait, upload flow etc)?

vsolovei-smartling:
> I'm confused, I thought switching to batch api v2 it's only about introducing new sdk dependency + rewriting upload queue handler/worker. Why are there so many changes to core code (massive refactoring of core trait, upload flow etc)?

The previous implementation relied on jobUids, vs batchUids for the submission upload queue in this one. I've reverted changes to `JobEntityWithBatchUid` but had to include the required changes from the previous implementation, as trying to merge into previous implementation produces a bigger changeset

PavelLoparev (approved):

inc/Smartling/Models/UploadQueueEntity.php
```diff
@@ -0,0 +1,67 @@
...
+
+    public static function getFieldLabels(): array
+    {
+        return [];
+    }
+
+    public static function getFieldDefinitions(): array
+    {
+        return [
+            self::FIELD_SUBMISSION_ID => SmartlingEntityAbstract::DB_TYPE_U_BIGINT,
```
dimitrystd, line 37: This DB record should contain one "command" - upload postX to 5 locales. Since each locale is represented by its own submission, then the table must reflect it. One db record must contain an array of submission ids.

dimitrystd (changes requested):
Left only one comment about DB table. I think it's the most important part of this refactoring. The queue describes the command for post/taxonomy. But each post is represented by one or many submissions + batchUid.

inc/Smartling/Models/UploadQueueEntity.php
vsolovei-smartling: Done

PR #526, opened 2024-04-22
add import flow (WP-826)

vsolovei-smartling:


inc/Smartling/ContentProvider.php
```diff
@@ -0,0 +1,158 @@
...
+        Content $assetUid,
+        $limit = 10,
+        $essential = false,
+        $childDepth = 0,
+        $relatedDepth = 0,
+        $found = 0,
+    ) {
+        $arguments = [
+            'numberposts' => $limit,
+            'parent_id' => $assetUid->getId(),
```
PavelLoparev, line 97: `$assetUid->getId()`, let's stick to uid (`getUid()`) naming?

inc/Smartling/RestApi.php
```diff
@@ -2,70 +2,177 @@
...
+            ]);
+            register_rest_route(self::NAMESPACE, 'assets/' . self::ASSET_UID_REGEX . '/details', [
                 'methods' => 'GET',
-                'permission_callback' => [$this, 'permissionCallback'],
+                'callback' => function (\WP_REST_Request $request) {
+                    $request = $this->setPrettyPrint($request);
+                    $assetUid = $this->getAssetUid($request->get_param('id'));
+                    if ($assetUid instanceof \WP_REST_Response) {
+                        return $assetUid;
+                    }
```
PavelLoparev, line 82: What does this mean? Response should be only given by corresponding method from `$this->getProvider()` object, why `$assetUid` can already be an instance of `WP_REST_Response`? And also, this code is repeated in every controller, maybe extract it to some method?

inc/Smartling/RestApi.php
```diff
@@ -2,70 +2,177 @@
...
+                    } catch (\JsonException $e) {
+                        return new \WP_REST_Response($e->getMessage(), 400);
+                    }
+
+                    return $this->getProvider()->getRelatedAssets(
+                        $assetUid,
+                        $parsed['limit'],
+                        $parsed['essential']['include'],
+                        ($parsed['child']['include'] ?? false) ? $parsed['child']['depth'] ?? 0 : 0,
+                        ($parsed['related']['include'] ?? false) ? $parsed['related']['depth'] ?? 0 : 0,
```
PavelLoparev, line 106: Let's introduce validation for incoming parameters? For now it seems like there is no validation rules for input at all? `$parsed` can be anything

inc/Smartling/RestApi.php
```diff
@@ -2,70 +2,177 @@
...
+        try {
+            $parts = explode('-', $string);
+            if (count($parts) < 2) {
+                throw new \InvalidArgumentException('AssetUid string expected to be contentType-id');
+            }
+            $id = array_pop($parts);
+
+            return new Content($id, implode('-', $parts));
+        } catch (\InvalidArgumentException $e) {
+            return new \WP_REST_Response($e->getMessage(), 400);
```
PavelLoparev, line 156: It's strange that business logic code returns infrastructure level objects (getAssetUid returns data or returns http response). I think returning http response should be implemented in place where this method is called.

PavelLoparev (changes requested):

inc/Smartling/ContentProvider.php
vsolovei-smartling: This is WP content id, don't think `getUid()` naming reflects this

inc/Smartling/RestApi.php
vsolovei-smartling: Refactored

inc/Smartling/RestApi.php
vsolovei-smartling: Done

inc/Smartling/RestApi.php
vsolovei-smartling: Refactored

PavelLoparev (approved):

PR #530, opened 2024-05-30
fix target content not created (WP-889)

vsolovei-smartling:


Buildplan/Dockerfile-Jenkins
```diff
@@ -75,10 +75,9 @@ RUN service mysql start && cd ${WP_INSTALL_DIR} && \
     ${WPCLI} core multisite-convert && \
     IFS=',' read -a array <<< "${SITES}"; for site in "${array[@]}"; do ${WPCLI} site create --slug="${site##*\:}" --title="${site%%\:*}" --email=test@wp.org; done && \
     IFS=';' read -a array <<< "${DEPENDENCIES}"; for plugin_name in "${array[@]}"; do ${WPCLI} plugin install $plugin_name --activate-network; done && \
-    if [ "latest" != "${ACF_PRO_VERSION}" ]; then DLINK="${DLINK}&t=${ACF_PRO_VERSION}"; fi && \
-    DLINK="https://connect.advancedcustomfields.com/index.php?a=download&p=pro&k=${ACFPRO_KEY}" && \
+    DLINK="https://smartling-connectors-storage.s3.amazonaws.com/wordpress/acf_pro/advanced-custom-fields-pro.5.11.zip" && \
```
vsolovei-smartling, line 78: Not issue related, but prevented builds in ci. The link now has an additional dynamic parameter when copied from the acf account page, and without it 403 status is returned

inc/Smartling/Base/SmartlingCoreUploadTrait.php
```diff
@@ -419,17 +419,22 @@ public function applyXML(SubmissionEntity $submission, string $xml, XmlHelper $x
     public function bulkSubmit(UploadQueueItem $item): void
     {
         $locales = $item->getSmartlingLocales()->getList();
-        $profile = $this->getSettingsManager()->getSingleSettingsProfile($item->getSubmissions()[0]->getSourceBlogId());
+        $profile = null;
```
dimitrystd, line 422: I don't see benefits from this change. Perhaps you prevented error in this line (when there is no submission), but I see `LiveNotificationController::pushNotification` that also expect not null `$profile`

inc/Smartling/DbAl/UploadQueueManager.php
```diff
@@ -35,7 +35,9 @@ public function __construct(
 
     public function count(): int
     {
-        return (int)$this->db->getRowArray("SELECT COUNT(*) cnt from $this->tableName")['cnt'];
+        return (int)$this->db->getRowArray(
+            "SELECT SUM(LENGTH(submission_ids) - LENGTH(REPLACE(submission_ids, ',', '')) + 1) cnt from $this->tableName"
```
dimitrystd, line 39: 🤯 🫤

inc/Smartling/Base/SmartlingCoreUploadTrait.php
dimitrystd: Perhaps you should remove `if ($profile === null)` in the loop below.

vsolovei-smartling: Added check for null profile in `pushNotification`.
I want get profile only once, so `if ($profile === null)` stays.

inc/Smartling/DbAl/UploadQueueManager.php
vsolovei-smartling: As we now store the submissionIds in a comma separated string, it's possible to count submissions per db row by calculating the count of commas in a column

PavelLoparev: we store array of values as a string? what is the data type of this column?

inc/Smartling/Base/SmartlingCoreUploadTrait.php
dimitrystd: You added condition inside the `catch`. But you also usage right below the loop
```
            LiveNotificationController::pushNotification(
                $profile->getProjectId(),
                LiveNotificationController::getContentId($submission),
                LiveNotificationController::SEVERITY_SUCCESS,
                vsprintf('<p>Sending file %s for locales %s.</p>', [
                    $submission->getFileUri(),
                    implode(',', array_values($locales)),
                ])
            );
```
I would recommend to refactor:
* Move `$profile = null;` and `$xml = '';` inside the `try`
* Get profile in the `catch` from submission again. 

It is much easy to read code when you see variable usage close to declaration.

vsolovei-smartling: I disagree with you on that one: I prefer try/catch blocks as small as possible.
Refactored to reduce lines changed

dimitrystd (approved):

PR #534, opened 2024-07-11
force register post type (WP-892)

vsolovei-smartling:
Since it's not possible to determine the reason of post type some times not being registered, we'll try to force register post type handler.
This must not be released for general availability.


dimitrystd (approved):
Good idea

PavelLoparev (approved):

vsolovei-smartling:
Action taken by client, this was not meant for public release

PR #541, opened 2024-09-11
fix batch creation error (WP-907)

vsolovei-smartling:
`array_unique()`, when removing an element that is not the last from an array turns a list (json encoded as `['a', 'b']`) into a map (json encoded as `{"0":"a","2":"b"}`)


inc/Smartling/Services/ContentRelationsDiscoveryService.php
```diff
@@ -302,7 +302,7 @@ public function createSubmissions(UserTranslationRequest $request): void
         $this->uploadQueueManager->enqueue($queueIds, $this->apiWrapper->createBatch(
             $profile,
             $jobInfo->getJobUid(),
-            array_unique($fileUris),
+            array_values(array_unique($fileUris)),
```
PavelLoparev, line 305: What was the issue? Array contained empty values?

PavelLoparev (approved):

dimitrystd (approved):

PR #548, opened 2024-11-05
remove log (WP-914)

vsolovei-smartling:
Log introduced temporarily, and is no longer needed

### DIFF
--- a/inc/Smartling/Bootstrap.php
+++ b/inc/Smartling/Bootstrap.php
@@ -435,7 +435,6 @@ class Bootstrap
          * Post types and taxonomies are registered on 'init' hook, but this code is executed on 'plugins_loaded' hook,
          * so we need to postpone dynamic handlers execution
          */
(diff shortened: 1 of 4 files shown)


inc/Smartling/Bootstrap.php
```diff
@@ -435,7 +435,6 @@ private function initializeBuildInContentTypes(ContainerBuilder $di): void
          * Post types and taxonomies are registered on 'init' hook, but this code is executed on 'plugins_loaded' hook,
          * so we need to postpone dynamic handlers execution
          */
-        self::$loggerInstance->info('Smartling bootstrap, action: ' . current_action());
```
PavelLoparev, line 438: I wanted to mention this on the last PR but I was like "nah, let it be here" :)

PavelLoparev (approved):

inc/Smartling/Bootstrap.php
vsolovei-smartling: The log served it's purpose, and was helpful though not related to the issue in that PR

PR #549, opened 2024-11-15
add option to skip ACF filter removal (WP-913)

vsolovei-smartling:
Escaping and slashing in the connector is fine.
The issue seems to be similar to an unsolved post from a long ago (https://support.advancedcustomfields.com/forums/topic/backslashes-being-added-to-page-post-content/)
This option might not solve the issue, but we'll have more info for the next iteration.

### DIFF
--- a/inc/Smartling/DbAl/WordpressContentEntities/PostEntityStd.php
+++ b/inc/Smartling/DbAl/WordpressContentEntities/PostEntityStd.php
@@ -203,8 +203,14 @@ class PostEntityStd extends EntityAbstract implements EntityWithPostStatus, Enti
         }
         $array = $entity->toArray();
         $array['post_category'] = \wp_get_post_categories($entity->ID);
-        // ACF would replace our properly escaped content with its own escaping.
-        remove_action('content_save_pre', 'acf_parse_save_blocks', 5);
+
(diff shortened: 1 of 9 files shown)


inc/Smartling/DbAl/WordpressContentEntities/PostEntityStd.php
```diff
@@ -203,8 +203,14 @@ public function set(Entity $entity): int
         }
         $array = $entity->toArray();
         $array['post_category'] = \wp_get_post_categories($entity->ID);
-        // ACF would replace our properly escaped content with its own escaping.
-        remove_action('content_save_pre', 'acf_parse_save_blocks', 5);
+
+        if (GlobalSettingsManager::isRemoveAcfParseSaveBlocksFilter()) {
+            // ACF would replace our properly escaped content with its own escaping.
+            $result = remove_action('content_save_pre', 'acf_parse_save_blocks', 5);
```
vsolovei-smartling, line 209: @dimitrystd, we've discussed re-adding this action after the function is done, but it makes no sense, because all content altering hooks are ran when `wp_insert_post` is called. Re-adding the function after calling the function is redundant.

dimitrystd (approved):
Approved, but we discussed to restore ACF action.

PR #550, opened 2024-11-22
Don't translate built in style strings (WP-911)

vsolovei-smartling:


tests/IntegrationTests/tests/GutenbergBlocksTest.php
```diff
@@ -82,23 +82,23 @@ public function testAttributesLocking()
...
 <!-- /wp:column --></div>
 <!-- /wp:columns -->
 
-<!-- wp:html {"fontSize":"[l~árgé]","otherAttribute":"[ó~thé~rVá~lúé]","smartlingLockId":"rootunlocked"} -->
+<!-- wp:html {"fontsize":"[l~árgé]","otherAttribute":"[ó~thé~rVá~lúé]","smartlingLockId":"rootunlocked"} -->
 <h1>[:~)]</h1>
 <!-- /wp:html -->
 
-<!-- wp:paragraph {"fontSize":"large","smartlingLockId":"rootlocked","smartlingLockedAttributes":"fontSize"} -->
+<!-- wp:paragraph {"fontsize":"large","smartlingLockId":"rootlocked","smartlingLockedAttributes":"fontSize"} -->
```
PavelLoparev, line 101: We only do not translate `style`, but `fontsize` we do, right? Do we have tests for `style` strings translation like these ones with `fontsize`?

vsolovei-smartling: `fontSize` is a built-in property, as is `style`, they should be copied. I had to rename the test property to `fontsize`, because it tests locking to ensure only locked properties get copied, other properties should translate

vsolovei-smartling: No tests currently test for `style` property

PavelLoparev: > No tests currently test for `style` property

ok, I thought this PR for `style` attribute.

But I saw test where `fontsize` still has pseudo translation. Is this expected in test? sounds like it should not happen

vsolovei-smartling: The filters are case sensitive, `fontSize` should now get copied, `fontsize` is translated. In this test, on lines 64-77 we update the attribute and add locking, changing the original translation to a user-modified string. The expected result for the test is that on next download, the replaced strings without locking to revert to pseudo-translation, and the locked ones should stay as the user has left them

tests/IntegrationTests/tests/GutenbergBlocksTest.php
```diff
@@ -33,23 +33,23 @@ public function testAttributesLocking()
         $postId = $this->createPost(title: 'Block attributes locking test', content: <<<HTML
 <!-- wp:columns {"smartlingLockId":"columns"} -->
 <div class="wp-block-columns"><!-- wp:column {"smartlingLockId":"leftcolumn"} -->
-<div class="wp-block-column"><!-- wp:paragraph {"fontSize":"large","smartlingLockId":"leftparagraph"} -->
+<div class="wp-block-column"><!-- wp:paragraph {"fontsize":"large","smartlingLockId":"leftparagraph"} -->
```
dimitrystd, line 36: Could you pls explain this change. You added configuration for `fontSize` but changed attribute here to keep test green. Is it correct? If so, then this approach is controversial. A different case is confusing, you may simply not notice it.

vsolovei-smartling: This is correct, renamed test attribute.

dimitrystd (approved):

PR #551, opened 2024-12-19
add exists check (WP-919)

vsolovei-smartling:


inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
```diff
@@ -261,8 +261,10 @@ protected function extractGroupsDefinitions(array $groups): array
         foreach ($groups as $group) {
             $defs[$group['key']] = [
                 'global_type' => 'group',
-                'active'      => $group['active'],
             ];
+            if (array_key_exists('active', $group)) {
+                $defs[$group['key']]['active'] = $group['active'];
+            }
```
PavelLoparev, line 267: Is it ok that acf group definition might have no `active` parameter? Maybe `'active' => FALSE` by default and then read value from `$group['active']`?

vsolovei-smartling: I've checked our code, and it looks like we never check this attribute anyway, it's a candidate for complete removal, after I confirm it

dimitrystd (approved):

PavelLoparev (approved):

PR #554, opened 2025-02-18
clear is_cloned flag on translation request (WP-927)

vsolovei-smartling:


readme.txt
```diff
@@ -62,6 +62,9 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
+= 3.10.0 =
+* Changed submission cloned status, now set to false when translating a previously cloned submission
```
dimitrystd, line 66: Reworded. "set to false" sounds too technical.
```suggestion
* Remove a cloned status when the user requests translation of the previously cloned submission
```

dimitrystd (approved):
Well, it is a questinable change. If I have a mix of assets (cloned or not) and want to re-request ALL without filter then I may get a surprise.

PavelLoparev (commented):
Why do we need to reset this flag? What issue does this solve?

vsolovei-smartling:
> Why do we need to reset this flag? What issue does this solve? I mean I've read ticket `There is no way to clean cloned status from the UI.` but I don't get why this flag needs to be reset.

Sometimes customers accidentally clone content that needs translation later. The only way to revert this currently is to query the database.

PR #558, opened 2025-05-06
skip after post created hooks (WP-935)

vsolovei-smartling:


inc/Smartling/DbAl/WordpressContentEntities/PostEntityStd.php
```diff
@@ -233,7 +233,7 @@ public function set(Entity $entity): int
         ));
         $preSavePostContent = $array['post_content'];
         $this->getLogger()->debug("Base64Encoded post content: " . base64_encode($array['post_content']));
-        $res = wp_insert_post($array, true);
+        $res = wp_insert_post($array, true, false);
```
PavelLoparev, line 236: Don't some other plugins rely on additional hooks after post insert? I'm afraid this fixes current issue but introduces more in other places.

dimitrystd: I reviewed the ticket description, and it mentions array parameters. But this change doesn't add anything to the `$array`.
I tend to agree with Pavel that it is unclear how this will affect third-party plugins.

vsolovei-smartling: Aftert testing, I discovered that adding the array parameter didn't help, and there's no way to handle new post creation without the flag to be compatible. Made it a global expert setting

dimitrystd (approved):

PavelLoparev (approved):
Make sure this new value equals param's default value

PR #561, opened 2025-06-24
fix yoast seo variables ingestion (WP-941)

vsolovei-smartling:


inc/Smartling/Helpers/MetaFieldProcessor/ReferencedStdBasedContentProcessorAbstract.php
```diff
@@ -76,12 +76,15 @@ public function processFieldPostTranslation(SubmissionEntity $submission, $field
...
-        ]);
+        $attachment = null;
+        if (is_int($value)) {
+            $attachment = $this->submissionManager->findOne([
+                SubmissionEntity::FIELD_CONTENT_TYPE => $this->detectRealContentType($submission->getSourceBlogId(), $value),
+                SubmissionEntity::FIELD_SOURCE_BLOG_ID => $submission->getSourceBlogId(),
+                SubmissionEntity::FIELD_SOURCE_ID => (int)$value,
+                SubmissionEntity::FIELD_TARGET_BLOG_ID => $submission->getTargetBlogId(),
+            ]);
+        }
```
PavelLoparev, line 87: `$this->submissionManager->findOne` throws if value is not int? Maybe wrap in try/catch then?

but main question: why caller code passes value (source id) as non int? is it valid? because it looks like we are fixing consequences but not root cause.

vsolovei-smartling: Code is responsible for finding attachment id on target site and is called for every field being sent for translation from a WordPress hook, value is of `mixed` type

vsolovei-smartling:
@PavelLoparev , I had to revert the isCloned check, as, unfortunately, it's being relied upon. Hopefully, code still made simpler

PR #564, opened 2025-07-01
use Elementor Pro document handling instead of custom fix (WP-938)

vsolovei-smartling:


vsolovei-smartling:
> Do existing tests pass?

Good catch, I've rearranged code for backward compatibility

PR #570, opened 2025-07-22
clear Elementor built-in documents cache after saving meta (WP-938)

vsolovei-smartling:
The issue occurred in the following sequence of events:

1. our plugin saves the post content
2. Elementor hook processes content, looks at metadata (there is none), caches the template type internally as the first available type (loop-item)
3. our plugin saves metadata
4. if Elementor elements are present in template, our plugin calls Elementor after save hook, but the template type there is read from cache instead of metadata and persists

Refactored Elementor elements present check and added a wrapper to clear Elementor Documents_Manager internal cache

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
   "name": "smartling/wordpress-connector",
   "license": "GPL-2.0-or-later",
-  "version": "3.12.0",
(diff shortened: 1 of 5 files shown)


PavelLoparev (commented):
Maybe add tests?

PR #582, opened 2025-10-02
Use a built-in function to search possible attachment id first (WP-961)

vsolovei-smartling:


tests/Smartling/Helpers/AbsoluteLinkedAttachmentCoreHelperTest.php
```diff
@@ -41,15 +44,40 @@ public function testLookForDirectGuidEntryQuery()
...
+        $this->invokeMethod(new AbsoluteLinkedAttachmentCoreHelper(
+            $this->createMock(SmartlingCore::class),
+            $this->createMock(AcfDynamicSupport::class),
+            $this->createMock(SubmissionManager::class),
+            $this->createMock(WordpressFunctionProxyHelper::class),
+            $this->createMock(WordpressLinkHelper::class),
+        ), 'lookForDirectGuidEntry', [$url]);
+    }
+
+    public function testProxyFunctionGetsCalledWhenLookingUpPossibleAttachmentId()
```
dimitrystd, line 56: This new test only checks that the new `method('attachment_url_to_postid')` will be called. But the new code added a condition, so we have at least two flows now. It would be great to have tests
- The new function returns valid image id
- The new function returns 0 and we fallback to existing logic

inc/Smartling/Helpers/WordpressFunctionProxyHelper.php
```diff
@@ -21,6 +21,11 @@ public function add_filter()
         return add_filter(...func_get_args());
     }
 
+    public function attachment_url_to_postid()
```
dimitrystd, line 24: Why don't you specify the return type of the function?

readme.txt
```diff
@@ -62,6 +62,10 @@ Additional information on the Smartling Connector for WordPress can be found [he
 3. Track translation status within WordPress from the Submissions Board. View overall progress of submitted translation requests as well as resend updated content.
 
 == Changelog ==
+= 4.3.6 =
+* Improved replacement of attachment urls in translated content
```
dimitrystd, line 66: Why not be more specific.
```suggestion
* Improved replacement of attachment URLs in translated content by using WordPress built-in attachment_url_to_postid() function
```

inc/Smartling/Helpers/AbsoluteLinkedAttachmentCoreHelper.php
```diff
@@ -217,6 +217,10 @@ private function getPossibleAttachmentIdByLocalOriginalFile(string $localOrigina
     {
         if (true === FileHelper::testFile($localOriginalFile)) {
             $originalPathinfo = pathinfo($localOriginalFile);
+            $wpAttachmentUrlToPostId = $this->wordpressProxy->attachment_url_to_postid($url);
```
dimitrystd, line 220: Why do you use this function only after checking that the file is present on the disk? It seems you only need to know the postId. Just want to check if this call should be the first line in the function.

inc/Smartling/Helpers/WordpressFunctionProxyHelper.php
vsolovei-smartling: Consistent with existing proxy pattern, we don't want to check the arguments and the return type

inc/Smartling/Helpers/AbsoluteLinkedAttachmentCoreHelper.php
vsolovei-smartling: Can be done, but the function is called for every absolute url, and only attachments are present on the disk

inc/Smartling/Helpers/AbsoluteLinkedAttachmentCoreHelper.php
```diff
@@ -228,6 +232,7 @@ private function getPossibleAttachmentIdByLocalOriginalFile(string $localOrigina
             if (null === $possibleId) {
                 $this->getLogger()->info(vsprintf('No \'attachment\' found for url=%s', [$url]));
             }
+            $this->getLogger()->debug("No attachment_url_to_postid result for url=$url, possible_id=$possibleId");
```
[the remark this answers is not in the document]
vsolovei-smartling: As was the intent. We can later make a decision to remove the old custom code if all the attachments are detected properly

tests/Smartling/Helpers/AbsoluteLinkedAttachmentCoreHelperTest.php
vsolovei-smartling: Done, it's mostly a static call on a static call, but at least we can test for db interaction

PavelLoparev (approved):

PR #612, opened 2026-04-16
use previous version of symfony cache (WP-996)

vsolovei-smartling:
I've ran into a similar issue that is likely related to https://github.com/symfony/symfony/discussions/46477. Two fixes: fix-double-namespace now includes an attempt to copy the unprocessed file to their namespaced location. The real fix is downgrading the symfony cache package to the previous major version to avoid an unprocessable file in docker environments.

### DIFF
--- a/fix-double-namespace.php
+++ b/fix-double-namespace.php
@@ -42,3 +42,80 @@ do {
     echo "\nDone!\n";
     ++$pass;
 } while ($filesModified > 0);
+
+// Copy classmap files that are present in inc/third-party but absent from inc/lib.
+// The namespacer skips namespace-less files (e.g. ValueWrapper.php), leaving them missing
+// in inc/lib even though the generated autoloader still expects them.
+$libDir = __DIR__ . '/inc/lib';
+$thirdPartyDir = __DIR__ . '/inc/third-party';
+$prefix = 'smartling-connector-';
+
+echo "\n" . str_repeat('=', $columns) . "\n";
+echo "Checking for classmap files missing from inc/lib...\n";
+echo str_repeat('=', $columns) . "\n";
+
+$filesCopied = 0;
+$filesMissing = 0;
+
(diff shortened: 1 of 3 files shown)


fix-double-namespace.php
```diff
@@ -42,3 +42,80 @@
...
+$prefix = 'smartling-connector-';
+
+echo "\n" . str_repeat('=', $columns) . "\n";
+echo "Checking for classmap files missing from inc/lib...\n";
+echo str_repeat('=', $columns) . "\n";
+
+$filesCopied = 0;
+$filesMissing = 0;
+
+foreach (new RecursiveIteratorIterator(
```
ykorobov-smartling, line 60: this is iterator that iterates over iterators? looks like 'pimp my ride' :)

ykorobov-smartling (approved):

fix-double-namespace.php
```diff
@@ -42,3 +42,80 @@
     echo "\nDone!\n";
     ++$pass;
 } while ($filesModified > 0);
+
+// Copy classmap files that are present in inc/third-party but absent from inc/lib.
+// The namespacer skips namespace-less files (e.g. ValueWrapper.php), leaving them missing
+// in inc/lib even though the generated autoloader still expects them.
+$libDir = __DIR__ . '/inc/lib';
+$thirdPartyDir = __DIR__ . '/inc/third-party';
+$prefix = 'smartling-connector-';
```
PavelLoparev, line 51: How does this file work? is it some kind of "migration"? does it run only when user updates the plugin?

vsolovei-smartling: Yes, this runs before release https://github.com/Smartling/wordpress-localization-plugin/blob/ad3f08b3d015d0ff3a300cd578da99d7ffccb874/composer.json#L49

PavelLoparev (approved):

PR #615, opened 2026-05-11
user interface fixes (WP-1001)

vsolovei-smartling:


inc/Smartling/ApiWrapper.php
```diff
@@ -103,7 +103,7 @@ private function getAuthProvider(ConfigurationProfileEntity $profile): AuthToken
         return $authProvider;
     }
 
-    public function acquireLock(ConfigurationProfileEntity $profile, string $key, int $ttlSeconds): \DateTime
+    public function acquireLock(ConfigurationProfileEntity $profile, string $key, float $ttlSeconds): \DateTime
```
PavelLoparev, line 106: why float for seconds?

PavelLoparev (approved):

sl-mmuradov (approved):

inc/Smartling/ApiWrapper.php
vsolovei-smartling: Locks are short lived, but not short enough lived to be in milliseconds. The change fixes an issue no one noticed or complained about, where .001 in test lock function was silently converted to 1.

PavelLoparev: sorry, I still do not get. [lock api accepts ttl in ms](https://smartling.atlassian.net/wiki/spaces/DEV/pages/4440677/Lock+Service+REST+API#Acquire-lock). you do not need to use float to describe fraction of a second

PR #616, opened 2026-05-14
rework ACF fields processing (WP-991)

vsolovei-smartling:


inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
```diff
@@ -108,237 +94,111 @@ private function getBlogListForSearch(): array
...
-                $defs = array_merge($defs, $this->extractGroupsDefinitions($local->groups));
-                $defs = array_merge($defs, $this->extractFieldDefinitions($local->fields));
-
+        if ($depth > self::MAX_ACF_FIELD_DEPTH) {
+            $this->getLogger()->error(sprintf(
+                'ACF field tree exceeded depth limit %d at field key "%s"; aborting recursion.',
+                self::MAX_ACF_FIELD_DEPTH,
+                $field['key'],
+            ));
+            return;
```
sl-mmuradov, line 162: Should it return? Can it just clamp the value?

inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
```diff
@@ -477,6 +322,13 @@ public function run(): void
         $this->tryRegisterACF();
     }
 
+    public function runIfRequired(): void
+    {
+        if ($this->definitions === null) {
+            $this->run();
```
sl-mmuradov, line 330: Looks weird. Hope it is done this way intentionally )

inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
```diff
@@ -535,19 +387,12 @@ public function getReplacerIdForField(array $attributes, string $key): ?string
         return null;
     }
 
-    public function getReferencedTypeByKey($key): string
+    public function getReferencedTypeByKey(string $key): string
     {
         if ($this->definitions === null) {
```
sl-mmuradov: Can we call `runIfRequired` here to reduce similar code?

inc/Smartling/Extensions/Acf/AcfTypeDetector.php
```diff
@@ -5,155 +5,97 @@
...
                 $fieldKey = $metadata['_' . $_realFieldName];
-                $this->getCache()->set($this->getCacheKeyByFieldName($fieldName), $fieldKey, static::$cacheExpireSec);
+                $this->cache->set($this->getCacheKeyByFieldName($fieldName), $fieldKey, self::CACHE_EXPIRE_SEC);
             } else {
-                return false;
+                return null;
             }
         }
 
         return $fieldKey;
```
sl-mmuradov, line 57: How did it work before? Was `$fieldKey` boolean? It is still boolean?

inc/Smartling/Extensions/Acf/AcfTypeDetector.php
```diff
@@ -5,155 +5,97 @@
...
-            return false;
+            return null;
         }
-        $conf = AcfDynamicSupport::$acfReverseDefinitionAction[$key];
-        $config = array_merge($conf, ['pattern' => vsprintf('^%s$', [$fieldName])]);
 
-        return CustomFieldFilterHandler::getProcessor(Bootstrap::getContainer(), $config);
+        $configuration['pattern'] = sprintf('^%s$', $fieldName);
+        $result = CustomFieldFilterHandler::getProcessor(Bootstrap::getContainer(), $configuration);
+        return $result ?: null;
```
sl-mmuradov, line 71: What is purpose to check return value for null here?

sl-mmuradov (approved):

PavelLoparev (changes requested):
### Issues

#### 🔴 Critical

**`inc/Smartling/WP/Table/QueueManagerTableWidget.php:276`** — Float `0.001` passed to `acquireLock()` which has `int $ttlSeconds` type hint.

Without `declare(strict_types=1)`, PHP silently coerces `0.001` → `0`, making the lock expire immediately — the duplicate-run guard never fires. The test covering this call path (`testPrepareItemsShowsRunningMessageWhenSdkWrapsGuzzle423`) was deleted in this same PR, so there is no safety net.

```php
// Fix:
$this->api->acquireLock($profile, JobAbstract::CRON_FLAG_PREFIX . $jobName, 1);
```

---

#### 🟡 Warning

**`inc/Smartling/Extensions/Acf/AcfDynamicSupport.php:141,174`** — `acf_get_fields()` returns `false` when a group has no fields; iterating `false` with `foreach` raises a PHP warning.

```php
// Fix:
foreach ((acf_get_fields($group) ?: []) as $field) {
```

**`inc/Smartling/Extensions/Acf/AcfDynamicSupport.php` — `loadDefinitions`** — If ACF is inactive, `$this->definitions` stays `null` after `tryRegisterACF`, so `runIfRequired()` re-enters `run()` on every subsequent call. Fix: assign `$this->definitions = []` in the `else` branch.

**`tests/Smartling/Extensions/Acf/AcfTypeDetectorTest.php:testGetProcessorForGutenberg`** — Test still sets the `$acf_stores` global which is now dead code after the refactor to `acf_get_field_groups()`. The test is fragile and its intent is unclear.

**`inc/Smartling/Helpers/RelativeLinkedAttachmentCoreHelper.php:63-65`** — The `count($this->acfDefinitions) === 0` guard is now indistinguishable between "not loaded" and "ACF has no fields," causing redundant `run()` calls. Use `runIfRequired()` instead.

**`inc/Smartling/Extensions/Acf/AcfDynamicSupport.php:addAcfFieldToDefs`** — Container fields with both a DB ID and inline `sub_fields` will have sub-fields processed twice (once from DB, once from inline array).

---

#### 🔵 Suggestion

**`inc/Smartling/ContentTypes/Elementor/ElementAbstract.php:113`** — `assert()` is disabled in production (`zend.assertions = -1`). Replace with an explicit `instanceof` check and log/exception.

**`inc/Smartling/DbAl/WordpressContentEntities/PostEntityStd.php:106`** — `get_edit_post_link()` default context changed from `'edit'` to `'display'` — different URL encoding. Verify intent.

**`inc/Smartling/Extensions/Acf/AcfDynamicSupport.php`** — `getDefinitions()` returns `[]` for both "not loaded" and "zero fields" — misleading API. Consider returning `null` when uninitialized.

---

### Recommendations

**Behavioral change — local ACF field groups**: The old code handled PHP/JSON-registered ACF groups via `$acf_local`/`$acf_stores`. The new code skips groups where `groupId <= 0`, meaning fields defined via `register_field_group()` without a DB post will no longer appear in `$definitions`. Needs product/QA sign-off.

**Test coverage gap**: `testPrepareItemsShowsRunningMessageWhenSdkWrapsGuzzle423` was deleted with no replacement. The behavioral change that motivated it makes sense, but the regression path is now uncovered.

**Strict types**: None of these files declare `strict_types=1` — the critical float-to-int coercion bug is a direct consequence.

---

### Assessment

**Ready to merge?** No — With fixes

**Reasoning:** The `float 0.001` silently coercing to `int 0` disables the distributed lock probe in `QueueManagerTableWidget`, which is the production guard against concurrent job runs — a silent behavioral regression that could cause duplicate job execution. The `acf_get_fields()` returning `false` without a null guard will also produce PHP warnings in any field group with zero children.

---
🤖 Review generated with [Claude Code](https://claude.ai/claude-code) (claude-pr-review-local skill)

inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
vsolovei-smartling: Not sure I've understand the clamping part, the return causes the function to stop any code below from running, so any deeper nested elements are not processed

inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
vsolovei-smartling: Done

inc/Smartling/Extensions/Acf/AcfDynamicSupport.php
vsolovei-smartling: Collecting definitions is an expensive action, so we don't load them every time plugin loads. External callers should not explicitly call run or runIfRequired. Reworked

inc/Smartling/Extensions/Acf/AcfTypeDetector.php
vsolovei-smartling: Simplified code

inc/Smartling/Extensions/Acf/AcfTypeDetector.php
vsolovei-smartling: CustomFieldFilterHandler::getProcessor return false when a processor is not found, and null is used in newer code. This function must return a null or MetaFieldProcessorInterface, returning false would result in a fatal error.

vsolovei-smartling:
> ### Issues
> #### 🔴 Critical
> **`inc/Smartling/WP/Table/QueueManagerTableWidget.php:276`** — Float `0.001` passed to `acquireLock()` which has `int $ttlSeconds` type hint.
Merged master into this branch

PavelLoparev (approved):

PR #619, opened 2026-05-21
fix content edit job view html (WP-1004)

vsolovei-smartling:


sl-mmuradov (approved):

tests/Smartling/WP/View/ContentEditJobViewTest.php
```diff
@@ -0,0 +1,80 @@
...
+            $hiddenOpens,
+            'Legacy display:none wrapper should still open (it hides the legacy .job-wizard fallback)'
+        );
+    }
+
+    private function resolveTemplate(bool $needWrapper): string
+    {
+        $source = $this->stripNonHtml(file_get_contents(self::VIEW_FILE));
+
+        $source = preg_replace(
```
PavelLoparev: Pattern 1 is dead on arrival: this PR removes the only `$needWrapper && false` block from the template, so this `preg_replace` will never match after merge. Consider removing it to avoid misleading future readers into thinking such a pattern might still exist in the template.

tests/Smartling/WP/View/ContentEditJobViewTest.php
```diff
@@ -0,0 +1,80 @@
...
+    public function testHtmlDivTagsBalanceOnPostEditScreen(): void
+    {
+        $html = $this->resolveTemplate(false);
+        $this->assertDivBalanced($html, 'post edit screen ($needWrapper=false)');
+    }
+
+    public function testHiddenWrapperHasClosingTag(): void
+    {
+        $source = $this->stripNonHtml(file_get_contents(self::VIEW_FILE));
+        $hiddenOpens = preg_match_all('#<div\s+style\s*=\s*"display:\s*none#i', $source);
```
PavelLoparev: The test name says "HasClosingTag" but the assertion only checks that at least one `<div style="display:none` *opening* tag exists — it never verifies a matching `</div>` closes it. The test passes trivially even if the closing tag is absent, which is exactly the class of bug this PR fixes.

tests/Smartling/WP/View/ContentEditJobViewTest.php
```diff
@@ -0,0 +1,80 @@
...
+        $this->assertGreaterThan(
+            0,
+            $hiddenOpens,
+            'Legacy display:none wrapper should still open (it hides the legacy .job-wizard fallback)'
+        );
+    }
+
+    private function resolveTemplate(bool $needWrapper): string
```
PavelLoparev: `resolveTemplate()` does not strip or simulate the outer `if (!$isBulkSubmitPage)` conditional, so the `#smartling-app` div is always included in the count regardless of context. A regression that only manifests on the bulk-submit screen (where that block is suppressed) would not be caught. Consider adding a `testHtmlDivTagsBalanceOnBulkSubmitScreen()` case that strips the `!$isBulkSubmitPage` block.

PavelLoparev (commented):
Code review findings on the new test file.

PavelLoparev (approved):

PR #621, opened 2026-05-28
add select all related assets ui (WP-1005)

vsolovei-smartling:
image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
   "name": "smartling/wordpress-connector",
   "license": "GPL-2.0-or-later",
-  "version": "5.5.1",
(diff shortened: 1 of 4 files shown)


sl-mmuradov (approved):

PavelLoparev (commented):
Why is 1st checkbox overlapped with buttons? can this be aligned?

vsolovei-smartling:
> Why is 1st checkbox overlapped with buttons? can this be aligned?

I've scrolled down a bit, but there still was a margin deficiency. Fixed, without scrolling it looks this way:
image

PavelLoparev (approved):

PR #626, opened 2026-07-02
add nonces (WP-1010)

vsolovei-smartling:


sl-mmuradov (approved):

PavelLoparev (commented):
do we need to do the same for other ajax requests?

PavelLoparev (approved):

vsolovei-smartling:
> do we need to do the same for other ajax requests?

No, they are ok

PR #628, opened 2026-07-13
add front end tests (WP-1011)

vsolovei-smartling:


PavelLoparev (approved):
## Review

**Overview**: Adds a Playwright E2E layer to catch regressions like WP-1007 (legacy jQuery AJAX call missing `_wpnonce`). Scope: CI infra (`Dockerfile-Jenkins`, `test.sh` E2E section: multisite domain normalization, custom static-file router, mu-plugin blocking outbound HTTP during tests, OPcache pre-warming, fixture creation), the Playwright suite itself (`auth.setup.js`, `ajax-security.spec.js`, `job-wizard.spec.js`), a WP-CLI fixture script, and a small `js/app.js` fix (swallow network errors in `loadRelations` instead of an unhandled rejection).

### Correctness / Test Coverage

1. **Bulk-submit-page nonce test can pass vacuously.** In `ajax-security.spec.js`, the post-edit-page test asserts `smartlingCalls.length > 0` before checking nonces; the bulk-submit-page version drops that assertion ("Bulk submit page may not trigger AJAX on load — only assert when calls exist"). But `js/app.js` (`JobWizard`'s `useEffect(() => { loadJobs(); }, [loadJobs])`) fires unconditionally regardless of `isBulkSubmitPage` — same `#smartling-app` container, same mount path. So if the bulk-submit page's AJAX call broke entirely (exactly the class of regression this PR targets), `smartlingCalls` and `missing` would both be `[]`, and `toHaveLength(0)` would pass with zero real coverage. Suggest mirroring the post-edit test's `expect(smartlingCalls.length).toBeGreaterThan(0)` guard here.

2. **Only half of WP-1007's DoD is covered.** WP-1007's DoD was "All ajax endpoints both verify nonces **and** check if user is capable of performing action." This suite verifies nonce presence / no-403s while authenticated as full admin, but doesn't test the capability-check side (no lower-privilege user scenario). Worth a follow-up ticket or a note in the PR description so reviewers don't assume that half is covered.

### Code Quality
- The new E2E section in `test.sh` is heavily and usefully commented given how much multisite/WordPress-bootstrap subtlety it's navigating — good for future maintainers.
- Nice cleanup: removed leftover `DIAG` debug echoes from the namespacer section.
- `js/app.js`'s new `catch (e) {}` is silent (comment-only), inconsistent with the sibling `loadJobs` catch which calls `setError(...)`. Probably fine since relations-loading is best-effort, but a `console.warn` would help future debugging.
- Minor: `Dockerfile-Jenkins` pipes NodeSource's setup script through `bash` unpinned — standard for CI images but worth knowing as an unpinned supply-chain point on every rebuild.

### Performance
- 900s timeout ceiling, OPcache pre-warming, static-file router bypassing WP bootstrap, and blocking outbound HTTP via mu-plugin are all sensible, well-justified mitigations for the CI failure modes described (clearly the product of the preceding "fix Playwright ..." commits).

### Security
- No production security issues. The HTTP-blocking/router/FLUpdater shims are CI-only and don't ship in the release zip (only `${PLUGIN_DIR}` is packaged).

### Suggested changes before merge
1. Add the missing `expect(smartlingCalls.length).toBeGreaterThan(0)` guard to the bulk-submit nonce test (or document why it's intentionally different).
2. Consider noting the WP-1007 capability-check gap in the PR description.

Nice addition overall — approving, with the above as non-blocking follow-ups.

